PR #5, opened 2022-12-19
Feature/handle multifile zip

EddyCMWF:


EddyCMWF:
@JamesVarndell , happy to revisit the implementation of this in the future, but at the moment handling zip files isn't working, and this is a handy little bug fix

PR #10, opened 2023-01-04
Makes geopandas an optional dependency

JamesVarndell:


EddyCMWF (approved):

EddyCMWF (changes requested):
import gbp inside __init__ method

EddyCMWF (approved):

PR #27, opened 2023-04-14
 let eccodes get raise if key does not exists

cosunae:
Description
============

The issue raised while testing the package with unstructured grid grib data, and calling to_xarray. 
Example to reproduce the error (with some ICON grib data): 
```
import earthkit.data
f_data="<...>"
fs = earthkit.data.from_source("file", f_data)
p_sel = fs.sel(param='P')
xarray_open_dataset_kwargs = {'backend_kwargs':{
                            "read_keys": ["typeOfLevel", "gridType"],
                        }, 'encode_cf':("time", "geography", "vertical")}
p_ds = p_sel.to_xarray(xarray_open_dataset_kwargs=xarray_open_dataset_kwargs)
```
The issue comes from the fact that that earthkit will codes_get with a default value, so that in case the grib file does not provide the key, the earthkit data object will still return None. 

However, cfgrib relies on codes_get raising (i.e. using codes_get w/o default) if the "latitudes" can not be decoded from the grib data (that is in the case of unstructured grid type). In that case, it will issue a warning but continues without lattitudes/longitudes coordinates. 
https://github.com/ecmwf/cfgrib/blob/66fba83176161aec33a01b8cbc086f2d4ce73d73/cfgrib/dataset.py#L426

If codes_get does not raise but returns a None default, it will try to build an ill-formed Variable with the array [None]. 

I am not familiar with the earthkit code base, so I am just opening this PR with the hope it is useful to demonstrate the issue. In case you would like to consider the PR for a fix, I am happy to provide tests if I am instructed what is the best way to do it (we would need to add probably additional grib data).


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=27) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=27) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/earthkit-data?pullRequest=27) it.</sub>

JamesVarndell:
Hi @cosunae,

Many thanks for your PR.

You raise a very valid point, but I'd like to discuss the possible implications of this change with some colleagues just in case something's relying on a default return value from the `get` method. We'll take a look next week and get back to you.

Thanks again,
James

cosunae:
great, thanks @JamesVarndell for looking into that. It makes sense, actually in my local env quite some tests do not pass, so I am pretty sure there are wider implications. As said, I can get more familiar with the tests and fix stuff. But first I wanted to get some hints from the developers if this is the right direction or not before investing more effort in a particular solution.

sandorkertesz:
Hi @cosunae,

Thank you for your pull request. It is a really useful fix but unfortunately, the proposed change has some implications because it breaks the GRIB metadata access in various parts of the code. To make it work some additional changes are needed. I can give you a set of instructions as to what needs to be added or I can push the changes straight into your PR (if you allow changes from maintainers). Please let me know which solution you prefer.  

Once it is all done and all the tests pass I suggest you add a test for your ICON grib data. I will give you the necessary details for doing that.

Thanks,
Sandor

cosunae:
Hi @sandorkertesz 
many thanks for the response and support. 
allow changes from maintainers is enabled so feel free to push changes. Either way works for me, probably is safer if you do it, since I am not yet too familiar with the code base. But I am happy to give it a try next week as well.
And if you instruct me on how to add grib data and a test, I am also happy to prepare it.

cosunae:
there seem to be a typo in the remote ? 
cosuane from cosuane/feature/fix_unstructured_grid2 should be cosunae ?

sandorkertesz:
Thank you! It works now and all the tests pass. Now the last task would be to add a test for converting your ICON data to xarray. I suggest you add this test to **tests/grib/test_grib_convert.py**. As for the input test data, I suggest one of these two options (supposing the file is called "test_icon.grib"):

- if it is very small (a few KB) and there are no copyright issues with it it can be added to the tests/data directory. Then you could load it with:
`f = from_source("file", earthkit_test_data_file("test_icon.grib")`
- host the data on the ECMWF download server and use it with: 
`f = from_source("url", earthkit_remote_test_data_file("examples", "test_icon.grib"))`
 In this case you need to share the data with me so that I can upload it to the server.

cosunae:
thanks @sandorkertesz, 
can you have a look at the data and test I added ? 
I double checked that the test is failing (cfgrib is raising the ValueError) with the main branch and fixed in this branch. 
I also merge with main to resolve a conflict. 
Let me know...

sandorkertesz:
Thanks @cosunae, It all look s good. Only the assert False should be removed from the test.

tests/grib/test_grib_convert.py
```diff
@@ -0,0 +1,25 @@
...
+def test_icon_to_xarray():
+    # test the conversion to xarray for an icon (unstructured grid) grib file.
+    g = from_source("file", earthkit_test_data_file("test_icon.grib"))
+    ds = g.to_xarray()
+    assert len(ds.data_vars) == 1
+    # Dataset contains 9 levels and 9 grid points per level
+    ref_levs = g.metadata("level")
+    assert ds["pres"].sizes["generalVerticalLayer"] == len(ref_levs)
+    assert ds["pres"].sizes["values"] == 6
+    assert False
```
sandorkertesz, line 25: This line should be removed

cosunae: upss yes, that shouldnt have been in the commit

sandorkertesz (approved):

PR #33, opened 2023-04-20
Adds CF grid mapping detection to get CRS information from netCDF input

JamesVarndell:


sandorkertesz:
Is this something that could be used for GRIB data too?

JamesVarndell:
> Is this something that could be used for GRIB data too?

Not directly, as this feature is specifically focused on understanding CF-compliant metadata. However, I have an open issue to develop something similar for GRIB.

JamesVarndell:
I have added a test skip for anything cartopy-related, so that if cartopy is not installed, the test will not run.

sandorkertesz:
Please can you change the target to the "develop" branch?

PR #35, opened 2023-04-21
Update environment.yml

samsammurphy:
1. jupyter is a dependency for the docs/examples
2. we should explicitly name the environment in the .yml?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,3 +1,4 @@
+name: earthkit-data
 channels:
 - conda-forge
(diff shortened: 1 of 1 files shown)


sandorkertesz:
There is an ongoing work to review the usage of environment.yml as part of #33, since it contains all the dependencies for testing, documentation etc. To install earthkit-data we need much less.

sandorkertesz (approved):

PR #44, opened 2023-05-03
pyfdb in environment.yml

samsammurphy:
[pyfdb](https://github.com/ecmwf/pyfdb) is required by [fdb_stream.ipynb](https://github.com/ecmwf/earthkit-data/blob/main/docs/examples/fdb_stream.ipynb).

It is not a conda package but it is on PyPI (i.e. can be installed using pip)

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -20,6 +20,7 @@ dependencies:
 - jupyterlab
 - pip:
   - git+https://github.com/ecmwf/multiurl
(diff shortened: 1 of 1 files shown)


sandorkertesz:
This is only needed for a notebook, which is skipped in the tests, so neither a test nor an installation requirement. To run pyfdb you would need additional binary dependencies.

PR #47, opened 2023-05-04
Split CI Teams notification into separate workflow

dtip:
So we can access the secret webhook even from PRs created from public forks.


dtip:
The `on-push` notify step fails because the new workflow isn't yet in the `main` branch. This is ready to merge.

sandorkertesz (approved):

PR #51, opened 2023-05-04
add link to eccodes grib namespace in grib_metadata.ipynb

samsammurphy:
Previous reference to ecCodes GRIB namespace

Screenshot 2023-05-04 at 14 38 52

is now a link to [this file](https://user-images.githubusercontent.com/16657983/236216946-9fea8461-1d43-4986-a258-2c81faee482c.png) and there is a statistics namespace example too (as GRIB namespaces might not be super familiar to some users)

Screenshot 2023-05-04 at 14 39 14

### DIFF
--- a/docs/examples/grib_metadata.ipynb
+++ b/docs/examples/grib_metadata.ipynb
@@ -3,7 +3,9 @@
   {
    "cell_type": "code",
    "execution_count": 1,
-   "metadata": {},
+   "metadata": {
+    "tags": []
+   },
    "outputs": [],
    "source": [
     "!test -f test6.grib || wget https://get.ecmwf.int/repository/test-data/emohawk/examples/test6.grib"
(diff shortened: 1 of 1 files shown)


sandorkertesz:
Nice improvement. I think the image should be stored somewhere else. Maybe on the ECMWF download server?

samsammurphy:
I agree with you @sandorkertesz, I will follow up offline on how to get the image into the ECMWF download server

sandorkertesz:
I just saw that this table was taken from the [ecCodes FAQ](https://confluence.ecmwf.int/display/UDOC/What+are+namespaces+-+ecCodes+GRIB+FAQ). What about simply adding a link to this page?

samsammurphy:
@sandorkertesz  your idea is better. I had (incorrectly) assumed that page was private. Just tested now in incognito mode and it still opens. Will update the PR to use the url to the [ecCodes FAQ](https://confluence.ecmwf.int/display/UDOC/What+are+namespaces+-+ecCodes+GRIB+FAQ)

PR #61, opened 2023-05-09
Switch to label-based CI approval for public PRs

dtip:
For a PR from a public fork, CI will only run when the "approved-for-ci"
label has been added after code review.
It will continue to run on push and (non-fork) PR as normal.


dtip:
The pre-commit hook failed to run for me so please squash this PR when merging to get rid of that commit which just ran the YAML autoformatter.

PR #69, opened 2023-05-10
Feature/review field api

### LINKED ISSUE #64: Review field API

sandorkertesz:
Brought GribField/FieldList methods in sync with the decisions of the API review discussions and added doctrings to them.

Fixes #64


sandorkertesz:
There is a read the docs page showing the GribField documentation:
https://earthkit-data.readthedocs.io/en/feature-review-field-api/_api/data/readers/grib/codes/index.html

Please carefully check it, especially the arguments used in ```metadata()``` and ```data()```.

iainrussell:
Is it possible to exclude the more 'internal' classes from the docs?

tlmquintino:
I dont mind that the internal classes are also in the docs, as long as we make **very clear** that they are internal details and should no be considered as part of the API.

sandorkertesz:
With sphinx-autoapi we can fine-tune what classes end up in the documentation. We just need to agree on what an internal class means and make sure non of the documented methods return internal classes. E.g. GribField has the public @handle property of the type CodesHandle. Now I think CodesHandle is an internal class, so the users should not have access to it. But if it is true @handle should be private too e.g. @_handle.

sandorkertesz:
The API docs has been greatly reduced, see: https://earthkit-data.readthedocs.io/en/feature-review-field-api/_api/data/readers/grib/codes/index.html

Issue #92, opened 2023-05-24
Should we add fdb as a separate source type?


sandorkertesz:
### Describe the solution you'd like

Currently to read from an fdb stream (with `pyfdb`) we need to do:

```
import pyfdb
import earthkit.data
r = {class': 'od', ...}
r = fdb.retrieve(request)
ds = earthkit.data.from_source("stream", r)
```
It came up during the discussions that we could use "fdb" itself as a source type. So we could write:

```
import earthkit.data
r = {class': 'od', ...}
ds = earthkit.data.from_source("fdb", request)
```

It seems to be a good idea, especially because earthkit-data already has the  "mars" and "cds" source types.

tlmquintino:
yes please. this would be very convinient.
we should have 2:
* fdb
* fdb stream

tlmquintino:
After discussing with @sandorkertesz, we believe its better to create one a **fdb** source and that _stream_ should become a method of the source, to be passed in the kwargs.

sandorkertesz:
Fixed by #93

PR #93, opened 2023-05-25
Feature/fdb

sandorkertesz:
A new source "fdb" was added. 

It is a very basic implementation. This source can be used as a **stream** or retrieve data into a (cached) **file**, controlled by the `stream=True` argument. See the "/docs/examples/grib_fdb_stream.ipynb" notebook example.

This code reads data from FDB as a stream:

```
request = {...}
ds = earthkit.data.from_source("fdb", request)
```

This code retrieves data from FDB into a file, which will be cached (the resulting `ds` is a fieldlist loaded from a file):

```
request = {...}
ds = earthkit.data.from_source("fdb", request, stream=False)
```
Related to #92

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,16 +101,16 @@ def _skip_for_api(app, what, name, obj, skip, options):
     # if what == "package":
     #     print(f"{name}[{what}]")
 
-    if what == "module" and "grib.codes" not in name and "grib.fieldset" not in name:
+    if what == "module" and "grib.codes" not in name and "grib.fieldlist" not in name:
         skip = True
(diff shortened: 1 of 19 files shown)


tlmquintino:
@sandorkertesz can you please remove the /home/fdbprod from the code.
For the moment we should throw an exception and ask the user to setup FDB_HOME and print also a link to the fdb readthedocs in the error message.

sandorkertesz:
@tlmquintino, I implemented the required changes for FDB_HOME. It is not set in the notebook example either.

tlmquintino (approved):
all good

PR #102, opened 2023-05-26
Add to_latlon to grib field

sandorkertesz:
1. `to_latlon()` returns a dict as:

   `{"lat": array, "lon": array}`

2. `to_points()` for these grid types `["regular_ll", "reduced_gg", "regular_gg"]` simply returns:

   `{"x": lon-array, "y": lat-array}`

    For other grid types it throws a ValueError. Not sure if it is the right implementation.

### DIFF
--- a/earthkit/data/readers/grib/codes.py
+++ b/earthkit/data/readers/grib/codes.py
@@ -456,6 +456,35 @@ class GribField(Base):
         return values
 
     def to_points(self, flatten=False):
+        r"""Returns the geographical coordinates in the data's original
+        Coordinate Reference System (CRS).
+
+        Parameters
+        ----------
+        flatten: bool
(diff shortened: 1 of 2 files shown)


earthkit/data/readers/grib/codes.py
```diff
@@ -456,6 +456,35 @@ def to_numpy(self, flatten=False, dtype=None):
...
+            y coordinates, respectively.
+
+        Raises
+        ------
+        ValueError
+            When the coordinates in the data's original CRS are not available.
+
+        """
+        grid_type = self.metadata("gridType", default=None)
+        if grid_type in ["regular_ll", "reduced_gg", "regular_gg"]:
```
tlmquintino, line 481: I think for the moment, until we have the grid library ready, this is good enough.

tlmquintino (approved):

PR #106, opened 2023-06-05
Feature/download examples

### LINKED ISSUE #52: Use python for sample file downloads in notebooks
The notebooks currently use bash to download sample files if not already on disk

Screenshot 2023-05-04 at 16 02 51

I really like this ^ one liner, however, users might not have `wget` installed on their machine necessarily. I say this because I was surprised to learn that I did not installed wget on my macbook yet (even if it is on every other machine I use).

Anywhos.. users will have python installed (presumably) if they are running this notebook, from the user's perspective downloading sample data might look something like 

Screenshot 2023-05-04 at 16 06 58

which is a bit easier to read. 

Here is an example python function (which I guess in practice will be a class method somewhere)...

```
import os
import urllib.request

def download_sample_file(filename):
    
    base_url = "https://get.ecmwf.int/repository/test-data/emohawk/examples/"
    data_url = base_url + filename
    
    if not os.path.isfile(filename):
        urllib.request.urlretrieve(data_url, filename)
```

sandorkertesz:
Fixes #52 

We have two new methods primarily used in the notebooks:

`download_example_file(filename, force=False)`: checks if the file is available and if it is not it downloads it (when `force=True` always downloads the file). Can take a list of filenames.
```
import earthkit.data
earthkit.data.download_example_file("temp_10.bufr")
fs = earthkit.data.from_source("file", "temp_10.bufr")
```

`remote_example_file`: build the url of the remote example file
```
import earthkit.data
from earthkit.data import remote_example_file
efas = earthkit.data.from_source("url", remote_example_file("efas.grib"))
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,11 +4,13 @@ on:
   push:
     branches:
     - main
+    - develop
     tags:
     - "*"
   pull_request:
     branches:
     - main
+    - develop
   pull_request_target:
     types: [labeled]
   workflow_dispatch:


iainrussell (approved):
Looks good! Better than the wget method.

PR #107, opened 2023-06-06
WIP: Add API tests for metadata handling

oiffrig:
A concrete example on how we could handle metadata. Still work in progress and open for discussion.

### DIFF
--- a/tests/metadata/test_metadata_api.py
+++ b/tests/metadata/test_metadata_api.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+# (C) Copyright 2020 ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+#
+
+import pytest
+
+from earthkit.data import from_source
+from earthkit.data.testing import earthkit_examples_file
+
+try:
+    from earthkit.data.metadata import Metadata
+except ImportError:
+    Metadata = None
+
+try:
+    from earthkit.data.metadata import RawMetadata
+except ImportError:
+    RawMetadata = None
+
+try:
+    from earthkit.data.readers.grib.metadata import GribMetadata
+except ImportError:
+    GribMetadata = None
+
+
+def test_abstract_metadata():
+    # Metadata properties:
+    # * immutable
+    # * behaves like a dict
+    # * can be updated (creating a new object, either a copy or a view)
+    # Updates can be made with compatible subclasses, e.g. update a GribMetadata with a RawMetadata
+    assert Metadata is not None
+
+
+def test_raw_metadata():
+    assert RawMetadata is not None
+    assert issubclass(RawMetadata, Metadata)
+
+    md = RawMetadata({"shortName": "2t", "perturbationNumber": 5})
+    assert md["shortName"] == "2t"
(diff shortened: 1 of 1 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=107) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=107) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/earthkit-data?pullRequest=107) it.</sub>

tests/metadata/test_metadata_api.py
```diff
@@ -0,0 +1,91 @@
...
+    assert GribMetadata is not None
+    assert issubclass(GribMetadata, Metadata)
+
+    with pytest.raises(TypeError):
+        GribMetadata({"shortName": "u", "typeOfLevel": "pl", "levelist": 1000})
+
+    f = from_source("file", earthkit_examples_file("test_single.grib"))
+    md = f[0].metadata()
+
+    assert isinstance(md, GribMetadata)
```
sandorkertesz, line 78: Do you suggest only to change the behaviour of `GribField.metadata()` when called without arguments, so, the rest of the functionality would be kept?

oiffrig: Yes, that was the idea. We keep the current API, but change the underlying metadata object, and we return it if the user is not asking for specific keys/namespaces

tests/metadata/test_metadata_api.py
```diff
@@ -0,0 +1,107 @@
...
+    assert RawMetadata is not None
+    assert issubclass(RawMetadata, Metadata)
+
+    md = RawMetadata({"shortName": "2t", "perturbationNumber": 5})
+    assert md["shortName"] == "2t"
+    assert md["perturbationNumber"] == 5
+    assert md.get("shortName") == "2t"
+    with pytest.raises(KeyError):
+        md["nonExistentKey"]
+    assert md.get("nonExistentKey", 12) == 12
```
sandorkertesz, line 53: What should happen for:

```python
md.get("nonExistentKey")
```
Should we raise `KeyError` or return None as a `dict` would do?

oiffrig: I would return None, to match `dict` and the behaviour of the default value when given

sandorkertesz: Thank you @oiffrig! There is another question related to the type cast. On a `Field` we can call `metadata()` like this to specify the value type we want:

```python
f.metadata("centre", astype=str)
f.metadata("centre", astype=int)
f.metadata("centre:s")
f.metadata("centre:str")
f.metadata("centre:l")
f.metadata("centre:int")
```
How would the Metadata object support these? The last 4 calls are of course supported for GRIB by `[]` and `get()`.

oiffrig: Not totally sure. What I would suggest is to have that as an overridable behaviour: the metadata object knows a few types and can convert from them (raising an error if the type is not supported), and specific implementations (e.g. GRIB) can extend the default behaviour by using functionalities of the underlying backend

sandorkertesz: Thanks. Would that mean that `get` should work like that:

```python
md.get("centre", astype=int)
```

sandorkertesz:
@oiffrig, I close this PR since all this was implemented in earthkit-data. Thank you for the contribution.

PR #118, opened 2023-06-15
Simplify usage of ls, describe and dump

sandorkertesz:
This PR simplifies the usage of the `ls`, `describe` and `dump` methods on GribFileld and BUFR objects as follows:

- in a notebook they behave as before, so a cell like this 

  ```python
  ds[0].dump()
  ```
  will generate an html output.

- the `print` option was removed. I we want to print their output we need to use the Python `print` command:

  ```python
  print(ds[0].dump())
  ```
- the `html` option was removed. If in a Jupyter notebook we need a non-HTML representation we just use the `print` command again:

  ```python
  print(ds[0].dump())
  ```
- `dump` now always returns a dict-like object, which in a notebook has an html representation while when printed with `print` behaves as dict


iainrussell (approved):
Looks good to me!

PR #122, opened 2023-06-19
Return list of BoundingBox objects on a FieldList

sandorkertesz:
This PR alters the way the `bounding_box` method used on a `FieldList` in line with the internal discussions regarding the FieldList API.  It also changes some of the method names of `BoundingBox`. The outcome is summarised below:

- the `bounding_box` method on a FieldList returns a list of `BoundingBox` objects (one per field, both for GRIB and NetCDF)
- `BoundingBox`  has a classmethod `union` generating the union of the input list of `BoundingBox` objects
- `BoundingBox`  has an instance method `union_with` generating the union with another object

```python
# getting bounding boxes on a FieldList
bb_list = ds.bounding_box()

# forming union
bb = BoundingBox.union(bb_list)

# forming union with the instance method
bb1 = BoundinBox(...)
bb2 = BoundingBox(...)
bb = bb1.union_with(bb2)
```

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -15,3 +15,9 @@ GRIB
 .. -----
 
 .. - :py:class:`BufrReader <data.readers.bufr.BUFRReader>`
+
+
+Other
+--------
+
+- :py:class:`BoundingBox <data.utils.bbox.BoundingBox>`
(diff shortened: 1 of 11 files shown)


iainrussell (approved):
This looks good to me Sandor!

PR #124, opened 2023-06-20
Implement bufr message and dump

sandorkertesz:
This PR is implementing the BUFR message dump that allows inspecting the hierarchical contents of a given BUFR message in a Jupyter notebook (as a tree view).
Since the dump is related to a given message the concept of a BUFR message object and BUFR message list was introduced.

The current state of the BUFR interface is summarised below:

A BUFR file is represented by a `BUFRMessageList` object, each ecCodes BUFR message is represented by a `BUFRMessage`. Iteration, slicing, selection and data extraction work:

```python
ds = earthkit.data.from_source("file", "temp_10.bufr")

# this gives the number of messages
len(ds)

for f in ds:
    print(f)

# this is one BUFRMessage
ds[0]

# lists metadata contents
ds.ls()

# extraction
df = ds.to_pandas(columns=("WMO_station_id", "data_datetime", "pressure", "airTemperature"))
```

Methods on a given message:

```python
# dumps message contents
ds[0].dump()

# dumps a given subset if the message contains subsets
ds[0].dump(subset=2)

# number of subests
ds[0].subset_count()

# nature of subsets
ds[0].is_compressed()
ds[0].is_uncompressed()
```

Concatenation/multiple files works
```python
ds1 = earthkit.data.from_source("file", "temp_a.bufr")
ds2 = earthkit.data.from_source("file", "temp_b.bufr")
ds = ds1 + ds2
ds.ls()
```

Slicing, sel:
```python
ds = earthkit.data.from_source("file", "temp_10.bufr")
ds[1:3]
ds.sel("edition=4)
```

Limitations:
- to make `to_pandas`work on multiple files a new release of pdbufr is needed (pdbufr code is in a feature branch at the moment)

Questions:
- the repr of a BUFRMessage is not satisfactory and has to be improved. It looks like this at the moment:
 `BUFRMessage(type=0,subType=1,subsets=1,20230602,120000)`


iainrussell:
This is looking really nice Sandor!

I have a half-baked thought - when I saw the method `subsets()` I thought that maybe it would return the subsets themselves as a new `BUFRMessageList`. Could that make sense, and would it be useful?

sandorkertesz:
> This is looking really nice Sandor!
> 
> I have a half-baked thought - when I saw the method `subsets()` I thought that maybe it would return the subsets themselves as a new `BUFRMessageList`. Could that make sense, and would it be useful?

I think it would be really useful, but would not implement it in this PR. I'd rather rename `subsets` to `subsets_count` to reserve the name for future use.

iainrussell:
Yes, that was really my thinking - `subsets()` sounds like it would return actual subsets, so we should have another name for a method that would return the count. `subset_count()` sounds good to me.

iainrussell:
This requires a new release of pdbufr, doesn't it?

sandorkertesz:
> This requires a new release of pdbufr, doesn't it?

Strictly speaking it would work with the current pdbufr release.  But `to_pandas` for multiple BUR files requires https://github.com/ecmwf/pdbufr/pull/61 to be released.

PR #126, opened 2023-06-26
Add control for number of bytes used for type checking in reader

### LINKED ISSUE #123: GRIB or BUFR files are recognised as unknown type
### What happened?

It happens when a GRIB or BUFR file does not start with the "GRIB" or "BUFR" bytes, but they contain some padding or data transmission headers. Then `from_source` assigns the `unknown` type to them.

### What are the steps to reproduce the bug?

Use 

```
 wget https://github.com/ecmwf/pdbufr/raw/master/tests/sample-data/M02-HIRS-HIRxxx1B-NA-1.0-20181122114854.000000000Z-20181122132602-1304602.bufr --output-document=hirs.bufr
```
then

```python
import earthkit.data
ds  = earthkit.data.from_source("file", "hirs.bufr")
```

### Version

any

### Platform (OS and architecture)

any

### Relevant log output

_No response_

### Accompanying data

_No response_

### Organisation

_No response_

sandorkertesz:
This PR fixes #123 and allows to correctly identify GRIB, BUFR and ODB data when they do not start with the "magic" bytes but contain some extra bytes at the beginning. As for NetCDF, this fix will not work because it must start with the four-byte “magic number” otherwise xarray, which is used to handle NetCDF in earthkit-data, cannot read it.

As a solution, the "reader-type-check-bytes" setting was added to control how many bytes the reader uses from the beginning of a source to determine its type. The default is 64 and the allowed range of values is 8 <= x < 4096. To make it work the concept of a Validator was introduced to Settings and the Interval object was implemented too.

We can use it like this when the default 64 bytes are not enough to identify the file type:

```python
from eathkit.data import from_source, settings

# set the number of bytes to 200. It will be saved into the setting file
settings.set("reader-type-check-bytes", 200):
ds = from_source("file", my_grib_with_extra_bytes_1)

# set the number of bytes to 256 temporarily
with settings.temporary("reader-type-check-bytes", 256):
    ds = from_source("file", my_grib_with_extra_bytes_2)


sandorkertesz:
> This looks good Sandor! There are lots of tests there that are unrelated to the code change, but I'm sure it's fine to keep them as part of the PR.

Unrelated test changes removed.

Issue #141, opened 2023-07-12
Add source type to access the Copernicus Atmosphere Data Store


sandorkertesz:
### Describe the solution you'd like

Add access to the Copernicus Atmosphere Data Store as a new source type "ads".

```python
ds = earthkit.data.from_source("ads", ....)
```

### Organisation

ECMWF

sandorkertesz:
Implemented with #142

PR #142, opened 2023-07-13
Feature/source ads

TomHodson:
Add the ability to use Atmosphere data store in addition to the climate data store. (https://github.com/ecmwf/earthkit-data/issues/141)

Mostly copied from [climetlab](https://github.com/ecmwf/climetlab/blob/develop/climetlab/sources/ads.py). The two APIs are almost identical save for having different endpoints and needing different authentication tokens.

The only change I made is a slight refactor to cds.py::CdsRetriever that allows us to subclass it for the AdsRetrriever rather than copying all that identical code. I also added some tests that I copied and adapted from tests/sources/test_cds.py

Let me know if there are any additional changes you'd like to see before merging this into earthkit. For instance CDSAPIKeyPrompt and ADSAPIKeyPrompt are also basically identical so could be merged if it seems worth it.

### DIFF
--- a/tests/sources/test_ads.py
+++ b/tests/sources/test_ads.py
@@ -10,14 +10,15 @@
 #
 
 import pytest
+import os
 
 from earthkit.data import from_source
-from earthkit.data.testing import NO_CDS
 
+NO_ADS = not os.path.exists(os.path.expanduser("~/.adsapirc"))
 
 @pytest.mark.long_test
 @pytest.mark.download
-@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
+@pytest.mark.skipif(NO_ADS, reason="No access to ADS")
 def test_ads_grib_1():
     s = from_source(
         "ads",
@@ -32,7 +33,7 @@ def test_ads_grib_1():
 
 @pytest.mark.long_test
 @pytest.mark.download
-@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
+@pytest.mark.skipif(NO_ADS, reason="No access to ADS")
 def test_ads_grib_2():
     s = from_source(
         "ads",
@@ -48,7 +49,7 @@ def test_ads_grib_2():
 
 @pytest.mark.long_test
 @pytest.mark.download
-@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
+@pytest.mark.skipif(NO_ADS, reason="No access to ADS")
 def test_ads_grib_3():
     s = from_source(
         "ads",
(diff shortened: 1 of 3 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=142) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=142) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/earthkit-data?pullRequest=142) it.</sub>

sandorkertesz:
The unit-tests fail because of #143

sandorkertesz:
Hi @TomHodson, please merge the latest develop into your branch to fix the failing test.

tests/sources/test_ads.py
```diff
@@ -0,0 +1,83 @@
...
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+#
+
+import pytest
+
+from earthkit.data import from_source
+from earthkit.data.testing import NO_CDS
```
sandorkertesz, line 15: `NO_CDS` is defined as:

```python
NO_CDS = not os.path.exists(os.path.expanduser("~/.cdsapirc"))
```

When we want to access both the CDS and the ADS via `cdsapi` we need to create a separate rc file for each. For example I use `~/.cdaspirc` for CDS and `~/.adsapirc` for ads. So using `NO_CDS` as defined here is not adequate to detect whether we have access to ADS. It can also happen that "~/.cdaspirc" contains the key for the ADS access! 

The bottom line is that, you need to define both `NO_CDS` and  `NO_ADS` with a more elaborate check on the availability of the given service.

tests/sources/test_ads.py
```diff
@@ -0,0 +1,86 @@
...
+# nor does it submit to any jurisdiction.
+#
+
+import os
+
+import pytest
+
+from earthkit.data import from_source
+
+NO_ADS = not os.path.exists(os.path.expanduser("~/.adsapirc"))
```
sandorkertesz, line 18: `NO_ADS` should be defined in `testing.py`, next to `NO_CDS`, so that other test fixtures can use it.

PR #148, opened 2023-07-17
134 add inputs transformer tools for earthkit subpackages

EddyCMWF:
Inputs transformer for use in sub-packages.
It uses type-setting, or explicit mapping, to try to ensure that the inputs passed to a function are converted to the appropriate type for that method. This means that earthkit users do not have to worry about the format of their data.

It is also possible for the user to bypass the conversion at the point of execution by setting the kwarg `_auto_convert_inputs = False`.

In the first instance this will be used earthkit-climate, but with further developments could/should be used by all earthkit subpackages


earthkit/data/utils/module_inputs_wrapper.py
```diff
@@ -0,0 +1,180 @@
...
+    Parameters
+    ----------
+    function : Callable
+        Method to be wrapped
+    kwarg_types : Dict[str: type]
+        Mapping of accepted object types for each arg/kwarg
+    convert_types : Tuple[type]
+        List of data-types to try to convert, this can be useful when the function is versitile and can
+        accept a large number of data-types, hence only a small number of types should be converted.
+
```
sandorkertesz, line 67: `**decorator_kwargs` is not documented.

earthkit/data/utils/module_inputs_wrapper.py
```diff
@@ -0,0 +1,180 @@
...
+                kwarg_type = T.get_args(kwarg_type)
+        else:
+            # 3. Do nothing, cannot assign None, as None is a valid type
+            continue
+        mapping[key] = kwarg_type
+    return mapping
+
+
+def transform_module_inputs(in_module, **decorator_kwargs):
+    """
```
sandorkertesz, line 163: This method is not tested.

earthkit/data/utils/module_inputs_wrapper.py
```diff
@@ -0,0 +1,180 @@
...
+
+def ensure_iterable(input_item):
+    """Ensure that an item is iterable"""
+    if not isinstance(input_item, (tuple, list, dict)):
+        return [input_item]
+    return input_item
+
+
+def ensure_tuple(input_item):
+    """Ensure that an item is iterable"""
```
sandorkertesz, line 42: Docstring should be corrected.

tests/utils/test_module_inputs_wrapper.py
```diff
@@ -0,0 +1,93 @@
...
+    assert WRAPPED_NP_MEAN(TEST_DA) == np.mean(TEST_NP)
+    assert WRAPPED_NP_MEAN(TEST_DA) == np.mean(TEST_DA)
+    assert isinstance(WRAPPED_NP_MEAN(TEST_DA), xr.DataArray)
+
+
+def test_transform_function_inputs_pandas_to_numpy():
+    # Test with pandas.DataFrame object
+    assert WRAPPED_NP_MEAN(TEST_DF) == np.mean(TEST_NP)
+    assert WRAPPED_NP_MEAN(TEST_DF) == np.mean(TEST_DF)
+    assert isinstance(WRAPPED_NP_MEAN(TEST_DF), np.float64)
```
sandorkertesz, line 93: I suggest you also add a test for the case when  `kwarg_types` or/and `convert_types` are invalid or do not match.

earthkit/data/utils/module_inputs_wrapper.py
EddyCMWF: fixed with SHA-fd58b85

earthkit/data/utils/module_inputs_wrapper.py
EddyCMWF: fixed with f9cd91e

earthkit/data/utils/module_inputs_wrapper.py
```diff
@@ -0,0 +1,180 @@
...
+        return tuple(ensure_iterable(input_item))
+    return input_item
+
+
+def transform_function_inputs(
+    function: T.Callable,
+    kwarg_types: T.Dict[str, T.Any] = {},
+    convert_types: T.Union[T.Tuple[T.Any], T.Dict[str, T.Tuple[T.Any]]] = (),
+    **decorator_kwargs
+) -> T.Callable:
```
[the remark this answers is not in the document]
EddyCMWF: I would say yes, as the transform_inputs approach is much more effective when type-setting is used by the functions which are being wrapped.

I do not want to remove this, but will do if it's a deal breaker.

I know type-setting is a controversial subject, but my position (and I think it is a reasonable one) is that type-setting is useful and should be used where possible and/or sensible but it should not be enforced, e.g. by mypy

tests/utils/test_module_inputs_wrapper.py
EddyCMWF: I'm not sure I understand the test you want.

If the input is not transformable to an expected type it is parsed unchanged the if that is invalid then the wrapped function/module would raise the error. Do you want me to include a test that asserts that an Error from the wrapped function is raised?

EddyCMWF: But to add to this, whilst adding the transform_module_inputs tests I ended up adding a lot more edge cases, which cover when we use type-setting of the wrapped methods and do not use kwarg_types/convert_types

sandorkertesz: > I'm not sure I understand the test you want.
> 
> If the input is not transformable to an expected type it is parsed unchanged, then if that is invalid then the wrapped function/module would raise the error. Do you want me to include a test that asserts that an Error from the wrapped function is raised?

Yes, what I suggest is to add some tests to see if the right exception is raised when the args are incorrect. .E.g. the following call should raise an exception since b is not an arg of np.mean():

```python
 module_inputs_wrapper.transform_function_inputs(
    np.mean,
    kwarg_types={"b": np.ndarray}
)
```

tests/utils/test_module_inputs.ipynb
sandorkertesz, line 1: What is the plan with this notebook?

EddyCMWF: sorry, commited inadvertedly, I have removed it

tests/utils/test_module_inputs_wrapper.py
EddyCMWF: This doesn't raise an exception, and I don't think it should (acknowledged this is a subjective opionion). Imposing a restraint here would mean that it would make it tricker to transform a series of functions with a single configuration (like I do with the transform_module_inputs)

tests/utils/test_module_inputs_wrapper.py
```diff
@@ -0,0 +1,169 @@
...
+    assert isinstance(WRAPPED_NP_MEAN(TEST_DA), xr.DataArray)
+
+
+def test_transform_function_inputs_pandas_to_numpy():
+    # Test with pandas.DataFrame object, axis=0 is to preserve the pandas.DataFrame, this is pandas syntax
+    result = WRAPPED_NP_MEAN(TEST_DF, axis=0)
+    assert result.equals(np.mean(TEST_DF, axis=0))
+    assert isinstance(result, pd.core.series.Series)
+
+    # Test without axis declartion, pandas>2 returns a scalar mean of the dataframe
```
sandorkertesz, line 166: Does it mean it would fail with pandas < 2?  There is a typo in "declartion".

PR #156, opened 2023-08-17
133 add pandas and geopandas wrappers

EddyCMWF:


JamesVarndell (approved):
Looks good to me - this will be a great addition for maximising ease of using shape-related tools!

PR #188, opened 2023-09-18
187 netcdf reader not initiated correctly

EddyCMWF:
I am still not exactly sure which circumstances trigger the bug, but I did find that it happened when handling level 2 satellite data (orbits). So I have included a test to ensure that level 2 data is readable. We can add more conditions when we know what methods we want available to this data object


tests/readers/test_netcdf_reader.py
```diff
@@ -221,6 +221,23 @@ def test_get_fields_missing_standard_name_attr_in_coord_array():
         assert len(fs) == 2
 
 
+def test_satellite_l2_data():
```
sandorkertesz, line 224: Please use the same pytest decorators as for the other cds-based tests. Something like this:

```python
from earthkit.data.testing import NO_CDS


@pytest.mark.long_test
@pytest.mark.download
@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
def test_satellite_l2_data():
```
Alternatively, you can add the data file to the download server and get it for the test. If it is very small it can be even added to `tests/data`.

earthkit/data/readers/netcdf.py
```diff
@@ -628,7 +628,7 @@ def mutate_source(self):
 class NetCDFReader(Reader):
     def __init__(self, source, path):
         Reader.__init__(self, source, path)
-        NetCDFFieldList.__init__(self, path)
```
sandorkertesz, line 631: `NetCDFFieldList` should be removed from this `__init__` since `NetCDFReader` has nothing to do with it!

tests/readers/test_netcdf_reader.py
EddyCMWF: I added the file to the test data server. Should I include any of the other decorators?

earthkit/data/readers/netcdf.py
EddyCMWF: I should delete the line entirely?

sandorkertesz: Yes, please.

EddyCMWF: done

tests/readers/test_netcdf_reader.py
EddyCMWF: I added the long_test and download decorators to my new test.
I also added some decorators to the other CDS test that was in there.

tests/readers/test_netcdf_reader.py
```diff
@@ -99,6 +101,8 @@ def test_dummy_netcdf_4():
 
 
 @pytest.mark.long_test
+@pytest.mark.download
+@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
 def test_netcdf_multi_cds():
     if not os.path.exists(os.path.expanduser("~/.cdsapirc")):
```
sandorkertesz, line 107: I you use NO_CDS this check is not needed. NO_CDS is doing the same:

```python
NO_CDS = not os.path.exists(os.path.expanduser("~/.cdsapirc"))
```

sandorkertesz (approved):

PR #195, opened 2023-09-19
194 write method added to NetCDFFieldList method

EddyCMWF:
write method added to NetCDFFieldList method, simply redirect to the to_netCDF method.

### DIFF
--- a/tests/sources/test_url.py
+++ b/tests/sources/test_url.py
@@ -136,7 +136,7 @@ def test_url_part_file_source():
 def test_url_netcdf_source_save():
     ek_url_nc_source = from_source(
         "url",
-        earthkit_remote_test_data_file("examples/efas.nc"),
+        earthkit_remote_test_data_file("examples/test.nc"),
     )
     ek_url_nc_source.save("temp.nc")
 
--- a/tests/sources/test_url.py
+++ b/tests/sources/test_url.py
@@ -15,7 +15,7 @@ import sys
 import pytest
 
 from earthkit.data import from_source, settings
-from earthkit.data.core.temporary import temp_directory
+from earthkit.data.core.temporary import temp_directory, temp_file
 from earthkit.data.testing import (
     earthkit_file,
     earthkit_remote_test_data_file,
@@ -134,11 +134,14 @@ def test_url_part_file_source():
 
 
 def test_url_netcdf_source_save():
-    ek_url_nc_source = from_source(
+    ds = from_source(
         "url",
         earthkit_remote_test_data_file("examples/test.nc"),
     )
-    ek_url_nc_source.save("temp.nc")
+
+    tmp = temp_file()
+    ds.save(tmp.path)
+    assert os.path.exists(tmp.path)
 
 
 if __name__ == "__main__":
--- a/tests/sources/test_url.py
+++ b/tests/sources/test_url.py
@@ -16,7 +16,11 @@ import pytest
 
 from earthkit.data import from_source, settings
 from earthkit.data.core.temporary import temp_directory
-from earthkit.data.testing import earthkit_file, network_off
+from earthkit.data.testing import (
+    earthkit_file,
+    earthkit_remote_test_data_file,
+    network_off,
+)
 
 
 @pytest.mark.skipif(  # TODO: fix
@@ -127,6 +131,14 @@ def test_url_part_file_source():
 
     with open(ds.path, "rb") as f:
         assert f.read() == b"GRIB7777GRIB7777"
+
+
+def test_url_netcdf_source_save():
+    ek_url_nc_source = from_source(
+        "url",
+        earthkit_remote_test_data_file("examples/efas.nc"),
+    )
+    ek_url_nc_source.save("temp.nc")
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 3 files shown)


tests/sources/test_url.py
```diff
@@ -129,6 +133,14 @@ def test_url_part_file_source():
         assert f.read() == b"GRIB7777GRIB7777"
 
 
+def test_url_netcdf_source_save():
```
sandorkertesz, line 136: This is a big file to download. It is enough if you just load the `test.nc` example (it also a fieldlist) as a "file" and save it.

sandorkertesz:
Please note that the `to_netcdf` method is not properly implemented, since it is saving the original netcdf data and ignoring the actual fieldlist structure.

EddyCMWF:
> Please note that the `to_netcdf` method is not properly implemented, since it is saving the original netcdf data and ignoring the actual fieldlist structure.

I think this is what we want, right? If we are just saving the object we want the original file where possible with minimal modification from EK. 

Actually I was a little unsure whether we should even be using xarray for opening and then writing, but I think that change can be part of another issue and PR.

sandorkertesz:
> > Please note that the `to_netcdf` method is not properly implemented, since it is saving the original netcdf data and ignoring the actual fieldlist structure.
> 
> I think this is what we want, right? If we are just saving the object we want the original file where possible with minimal modification from EK.
> 
> Actually I was a little unsure whether we should even be using xarray for opening and then writing, but I think that change can be part of another issue and PR.

Consider this code:

```python
# ds is a fieldlist with e.g. 100 fields
ds = earthkit.data.from_source("file", "my_data.nc")
f = ds[2:10]
f.save("my_file.nc")
```

EddyCMWF:
> > > Please note that the `to_netcdf` method is not properly implemented, since it is saving the original netcdf data and ignoring the actual fieldlist structure.
> > 
> > 
> > I think this is what we want, right? If we are just saving the object we want the original file where possible with minimal modification from EK.
> > Actually I was a little unsure whether we should even be using xarray for opening and then writing, but I think that change can be part of another issue and PR.
> 
> Please consider these examples:
> 
> ```python
> # ds is a fieldlist with e.g. 100 fields
> ds = earthkit.data.from_source("file", "my_data.nc")
> f = ds[2:10]
> f.save("my_file.nc")
> ```
> 
> ```python
> # ds1 is a fieldlist with 10 fields
> ds1 = earthkit.data.from_source("file", "my_data_1.nc")
> # ds2 is a fieldlist with 20 fields
> ds2 = earthkit.data.from_source("file", "my_data_2.nc")
> # concatenation
> ds = ds1 + ds2
> f = ds[8:12]
> f.save("my_file.nc")
> ```


I added a not implemented error to the NetCDFMaskedFieldList as this was producing an error.

The MultiFile looks to work if the 2 data objects are concatable.

sandorkertesz:
> > > > Please note that the `to_netcdf` method is not properly implemented, since it is saving the original netcdf data and ignoring the actual fieldlist structure.
> > > 
> > > 
> > > I think this is what we want, right? If we are just saving the object we want the original file where possible with minimal modification from EK.
> > > Actually I was a little unsure whether we should even be using xarray for opening and then writing, but I think that change can be part of another issue and PR.
> > 
> > 
> > Please consider these examples:
> > ```python
> > # ds is a fieldlist with e.g. 100 fields
> > ds = earthkit.data.from_source("file", "my_data.nc")
> > f = ds[2:10]
> > f.save("my_file.nc")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```python
> > # ds1 is a fieldlist with 10 fields
> > ds1 = earthkit.data.from_source("file", "my_data_1.nc")
> > # ds2 is a fieldlist with 20 fields
> > ds2 = earthkit.data.from_source("file", "my_data_2.nc")
> > # concatenation
> > ds = ds1 + ds2
> > f = ds[8:12]
> > f.save("my_file.nc")
> > ```
> 
> I added a not implemented error to the NetCDFMaskedFieldList as this was producing an error.
> 
> The MultiFile looks to work if the 2 data objects are concatable.

Please consider this code:

```python
# ds1 is a fieldlist with 10 fields
ds1 = earthkit.data.from_source("file", "my_data_1.nc")
# ds2 is a fieldlist with 20 fields
ds2 = earthkit.data.from_source("file", "my_data_2.nc")

f1 = ds1[2:5]
f2= ds2[8:14]
f = f1 + f2
 f.save("my_file.nc")
```

EddyCMWF:
> > The MultiFile looks to work if the 2 data objects are concatable.
> 
> Should be tested. But please consider this code:
> 
> ```python
> # ds1 is a fieldlist with 10 fields
> ds1 = earthkit.data.from_source("file", "my_data_1.nc")
> # ds2 is a fieldlist with 20 fields
> ds2 = earthkit.data.from_source("file", "my_data_2.nc")
> 
> f1 = ds1[2:5]
> f2 = ds2[8:14]
> f = f1 + f2
> f.save("my_file.nc")
> ```

But this is a problem with the to_xarray method of the NetCDFMaskedFieldList object, which I don't think is the responsibility of this issue and PR. I have added another not implemented exception for when the objects do not a path attribute.

EddyCMWF:
hey @sandorkertesz, thanks for finishing this off :)

PR #200, opened 2023-09-21
Add documentation for the wekeo sources

sandorkertesz:
See the documentation of the "wekeo" and "wekeocds" sources at: https://earthkit-data--200.org.readthedocs.build/en/200/guide/sources.html


iainrussell (changes requested):
Looks good to me - I think it's sensible enough to have the two wekeo sources as different sources. I might suggest removing the quotes around "grammar" - the quotes are not there in the example notebook and it looks fine there.

iainrussell (approved):
Wonderful, thank you! Looks like a good example of using each source.

Issue #180, opened 2023-09-15
The "file" source is not able to open with pandas CDS files previously downloaded


malmans2:
### What happened?

I'm able to open CDS files downloaded using `from_source("cds", ...)`, but I'm not able to open them if they've been previously downloaded (i.e., using `from_source("file", ...)`).

In this specific case, when using `from_source("cds", ...)` looks like additional arguments are passed to the libraries used under the hood to read data (pandas, `comment="#"`).

Is there a way to open a local file previously downloaded from the CDS exactly as `from_source("cds", ...)` would do?

### What are the steps to reproduce the bug?

```python
import cdsapi
import earthkit.data

collection_id = "insitu-observations-gruan-reference-network"
request = {
    "format": "csv-lev.zip",
    "year": "2006",
    "month": "05",
    "variable": ["air_temperature", "altitude"],
    "day": ["21", "22"],
}

data_cds = earthkit.data.from_source("cds", collection_id, **request)
data_cds.to_pandas()  # OK

data_file = earthkit.data.from_source("file", data_cds.path)
data_file.to_pandas()  # ParserError

client = cdsapi.Client()
data_cdsapi = earthkit.data.from_source(
    "file", client.retrieve(collection_id, request).download()
)
data_cdsapi.to_pandas()  # ParserError
```

### Version

0.3.1

### Platform (OS and architecture)

Darwin MacBook-Pro-3.local 22.6.0 Darwin Kernel Version 22.6.0: Wed Jul  5 22:21:56 PDT 2023; root:xnu-8796.141.3~6/RELEASE_X86_64 x86_64

### Relevant log output

```shell
---------------------------------------------------------------------------
ParserError                               Traceback (most recent call last)
Cell In[1], line 17
     14 data_cds.to_pandas()  # OK
     16 data_file = earthkit.data.from_source("file", data_cds.path)
---> 17 data_file.to_pandas()  # ParserError

File ~/mambaforge/envs/earthkit/lib/python3.10/site-packages/earthkit/data/sources/file.py:107, in FileSource.to_pandas(self, **kwargs)
    105 def to_pandas(self, **kwargs):
    106     LOG.debug("Calling reader.to_pandas %s", self)
--> 107     return self._reader.to_pandas(**kwargs)

File ~/mambaforge/envs/earthkit/lib/python3.10/site-packages/earthkit/data/readers/csv.py:144, in CSVReader.to_pandas(self, **kwargs)
[...]

PR #211, opened 2023-10-02
consistent pandas_read_csv_kwargs for file and CDS sources

EddyCMWF:
See #180 for full details and background.

CDS and file sources now have consistent pandas_read_csv_sources such that a resource from  theCDS has the same behaviour whether opened automatically or saved for a later date. Several other small refactorisations also included.

I have not yet modified the ecmwf_api default kwargs as that has more bespoke requirements. But this is due to them being TAB sperated values, not COMMA seperated values.


EddyCMWF:
Latest commit is following an offline discussion to improve the visibility and explicitness of the kwarg handling in the read_csv method

earthkit/data/readers/csv.py
```diff
@@ -134,13 +134,26 @@ def __init__(self, source, path, compression=None):
         self.compression = compression
         self.dialect, self.has_header = probe_csv(path, compression=compression)
 
-    def to_pandas(self, pandas_read_csv_kwargs=None):
+    def to_pandas(
+        self, comment="#", compression=None, pandas_read_csv_kwargs=None, **kwargs
```
sandorkertesz, line 138: Why do we need **kwargs here?

earthkit/data/sources/cds.py
```diff
@@ -123,27 +123,5 @@ def requests(self, **kwargs):
 
         return result
 
-    def to_pandas(self, **kwargs):
-        pandas_read_csv_kwargs = dict(
-            comment="#",
-            parse_dates=["report_timestamp"],
```
sandorkertesz, line 129: `parse_dates` are not added as an argument to csv.to_pandas()

tests/sources/test_cds.py
```diff
@@ -102,6 +102,24 @@ def test_cds_netcdf_selection_limited():
     assert len(s) == 9
 
 
+@pytest.mark.long_test
```
sandorkertesz, line 105: A similar test should be added to `ecmwf_api`

tests/readers/test_csv_reader.py
```diff
@@ -141,11 +141,11 @@ def test_csv_with_comment():
         comment_line="This is a comment",
     )
 
-    df = s.to_pandas(pandas_read_csv_kwargs={"comment": "#"})
```
sandorkertesz, line 144: We still need to add a test case that is explicitly using `pandas_read_csv_kwargs`. With this change there is none.

earthkit/data/readers/csv.py
sandorkertesz, line 138: We could also have: 
```python
to_pandas(self, pandas_read_csv_kwargs={"comment": "#", "compression": None})
```

earthkit/data/readers/csv.py
```diff
@@ -134,13 +134,26 @@ def __init__(self, source, path, compression=None):
...
+    ):
         """Convert CSV data into a :py:class:`pandas.DataFrame` using :py:func:`pandas.read_csv`.
 
         Parameters
         ----------
+        comment: str
+            Character that represents a comment line in csv file. This value is ignored if the comment
+            character is defined in pandas_read_csv_kwargs.
+        compression: (None, str)
+            The compression of the file being opened. This value is ignored if Earthkit detects
```
sandorkertesz, line 148: The user will not know if earthkit detected the compression, so this decision is hidden.

earthkit/data/readers/csv.py
EddyCMWF: This was to make the user API nicer such that if we are calling to pandas on a CSV reader you don't need to nest the kwargs, they can be passed in directly. The nesting is maintained such that when to_pandas is called by an intermediate method, the read_csv_kwargs can be parsed without interfering with intermediate method. Hopefully that makes sense, we can discuss in person if you want

tests/readers/test_csv_reader.py
EddyCMWF: added test: `test_csv_5`

earthkit/data/readers/csv.py
EddyCMWF: I agree, but this partly why I included it in the doc-string, so that it is clear that Earthkit is trying to do some compression detection. TBH, it probably makes most sense to remove this special treatment we give to compression as I think most out of the box cases the Earthkit detection is doing it all. I was just trying to be as compatible with the existing code.
... maybe this is another point to discuss in person?

earthkit/data/sources/cds.py
EddyCMWF: test added to test_cds.py: `test_cds_non_observation_csv_file`

earthkit/data/readers/csv.py
EddyCMWF: As discussed the `kwargs` are not used by this method. However, I have left them in the definition such that the code does not break if a user provides kwargs that are not applicable, e.g. the ODB kwargs.

tests/sources/test_cds.py
EddyCMWF: As discussed, tests added for other formats from the CDS (grib and netcdf).

earthkit/data/readers/csv.py
EddyCMWF: I have removed compression from the kwargs as I think it is misleading/confusing, earhtkit will overwrite any provided value with what it detects.

I have added a note in the docstring explaining the EK will handle the compreesion of archived csv files

tests/readers/test_csv_reader.py
```diff
@@ -89,6 +89,23 @@ def test_csv_4():
...
+        headers=["a", "b", "c"],
+        quote_strings=True,
+        lines=[
+            [1, "x", 3],
+            [4, "y", 6],
+            [7, "z", 9],
+        ],
+    )
+
+    df = s.to_pandas(pandas_read_csv_kwargs={"index_col": "a"})
```
EddyCMWF, line 105: Hi @sandorkertesz , here is my test of the pandas_read_csv_kwargs

sandorkertesz (approved):

EddyCMWF:
Can this be merged?

Issue #196, opened 2023-09-19
Attempt to detect filename when saving


EddyCMWF:
### Is your feature request related to a problem? Please describe.

The hidden file objects often have filenames that are long and logical, e.g. when downloading data from the CDS the designated filename generally describes the context of the file. It is not optimum to provide a filename everytime we save a file, particularly if we intend to use earthkit as the primary/supported method of downloading data from the CDS and ADS.

### Describe the solution you'd like

To detect the basename of the source object and use that as the download filename in the current directory.

Implement as a decorator for flexibility.

PR #218, opened 2023-10-12
196 attempt to detect filename when saving

EddyCMWF:
What do you think of this? Maybe one to discuss in an earthkit-data meeting?
The issue #196 has the back story


tests/translators/test_translators.py
```diff
@@ -133,7 +134,7 @@ def test_gpd_dataframe_translator():
 
 def test_transform_from_grib_file():
     # transform grib-based data object
-    f = from_source("file", "tests/data/test_single.grib")
+    f = from_source("file", earthkit_test_data_file("test_single.grib"))
```
EddyCMWF, line 137: Hi @sandorkertesz ,
I had to include this to get the tests to pass. It is probably the correct way to load files, but I am quite confused why this only fails on my PR as my code change should not touch these tests. Any ideas? Or are you happy for this improvement to the tests to be included in this PR.
Eddy

sandorkertesz: Yes,  `earthkit_test_data_file()` is right method to use here.

sandorkertesz:
Please can you add a comment showing some examples with explanation? #196 does not contain these.

earthkit/data/decorators.py
```diff
@@ -35,6 +35,19 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+def detect_out_filename(func):
+    @functools.wraps(func)
+    def wrapped(self, *args, **kwargs):
+        if len(args) == 0:
+            try:
+                args = [os.path.basename(self.path)]
```
sandorkertesz, line 43: What if I have this code:

```python
ds = from_source("file", "my.grib")
ds.save()
```
or
```python
ds = from_source("file", "my.grib")
ds.save(append=True)
```

EddyCMWF: image

EddyCMWF: ds.save() created a file called test6.grib in the current directory.

the `append` kwarg was not recognised with the decorator... just looking into why

EddyCMWF:
> Please can you add a comment showing some examples with explanation? #196 does not contain these.

This has raised a good point, it looks like we have already thrown away the real filename by the time we get to the decorator, so the filename used is the just the earthkit cache hash. I will also look into how we can handle this better.

earthkit/data/decorators.py
sandorkertesz: The first example overwrites the original file we just read with `from_source()`!!!
```python
ds = from_source("file", "my.grib")
ds.save()
```

The second the example would append to itself to the original file:
```python
ds = from_source("file", "my.grib")
ds.save(append=True)
```

Neither should happen!

EddyCMWF: but doing this would produce the same result in the current implementation:
```
ds = from_source("file", "my.grib")
ds.save("my.grib")
ds.save("my.grib", append=True)
```

sandorkertesz: In these examples `ds` is a not FileSource!

EddyCMWF: GribReader, sorry.

EddyCMWF: Actually that didn't fix the exisiting bug that it is possible to overwrite the file you are reading. The latest commit fixes this with a warning

earthkit/data/decorators.py
```diff
@@ -35,6 +35,26 @@ def wrapped(*args, **kwargs):
...
+            except AttributeError:
+                raise TypeError("Please provide a output filename")
+
+        if hasattr(self, "path") and args[0] == self.path:
```
sandorkertesz, line 47: This check is not enough. Consider this:
```python
ds = from_source("file", "./my.grib")
ds.save()
```
Now args[0]="my.grib" and self.path="./my_grib". So they are different but point to the same file.

earthkit/data/decorators.py
sandorkertesz: > GribReader, sorry.

Still the error says `TypeError` FileSource.save()` so something goes wrong.

earthkit/data/decorators.py
EddyCMWF: good point, I've change to os.path.samefile instead.

EddyCMWF: But given the other point on here, I think this may need to change again when I address that the self.path is the cache hash, not the source filename we want. So I am going to work on that angle before coming back to any other changes here.

EddyCMWF:
> > Please can you add a comment showing some examples with explanation? #196 does not contain these.
> 
> Here is the example I am working towards:
> 
> ```
> s = from_source(
>     "cds",
>     "reanalysis-era5-single-levels",
>     variable=["2t"],  #, "msl"],
>     product_type="reanalysis",
>     area=[50, -50, 20, 50],
>     date="2007-01-01/2008-02-04",
>     time="12:00",
> )
> s.save()
> ```
> 
> I will add to the cds notebook, and something similar in a URL notebook.
> 
> Working on this has raised a good point, it looks like we have already thrown away the real filename by the time we get to the decorator, so the filename used is the just the earthkit cache hash. I will also look into how we can handle this better.

earthkit/data/decorators.py
EddyCMWF: The error I posted above? In the end I realised that that was caused by what we have been discussing, when we saved we tried to overwrite the file we were reading, and this created a broken grib, so when I opened it again it was not understood as a grib. This is fixed by the check to not overwrite the existing file, but I am going to look into the preserving the source filename before re-opening this ticket.

EddyCMWF: This point has been addressed with the latest changes. I have added tests that ensure that the warning is raised when the user attempts to overwrite the file they are reading.

EddyCMWF:
Hi @sandorkertesz ,
I've reopened this PR, it now includes safety features which prevent users from over-writing the file they are reading, and we preserve the source_filename from the CDS.

I'm still not convinced by the behaviour when the result is an archive, but there are other deeper issues I have there, so I will create a new issue for that.

Eddy

earthkit/data/decorators.py
EddyCMWF: This is addressed by using os.path.samefile

earthkit/data/decorators.py
```diff
@@ -35,6 +36,38 @@ def wrapped(*args, **kwargs):
...
+    @functools.wraps(func)
+    def wrapped(self, *args, **kwargs):
+        # Detect filename:
+        if len(args) == 0:
+            for att in ["source_filename", "path"]:
+                if hasattr(self, att) and getattr(self, att) is not None:
+                    args = [os.path.basename(getattr(self, att))]
+                    break
+            else:
+                raise TypeError("Please provide a output filename")
```
sandorkertesz, line 49: Fix: "a output"

earthkit/data/readers/__init__.py
```diff
@@ -60,6 +65,7 @@ def ignore(self):
     def cache_file(self, *args, **kwargs):
         return self.source.cache_file(*args, **kwargs)
 
+    @detect_out_filename
```
sandorkertesz, line 68: Why we needf the decorator here? We cannot call it without an argument.

sandorkertesz:
> Hi @sandorkertesz , Is there anything blocking this being merged? Eddy

There are some pending comments. Otherwise, I just need to check it again, it has been a long time since I last looked into it

EddyCMWF:
Hi @sandorkertesz , I don't see any unresolved comments, could you point me to them?

sandorkertesz:
> Hi @sandorkertesz , I don't see any unresolved comments, could you point me to them? Can you see any of them?

E.g. this one (and the ones above it on the page)
image

sandorkertesz (changes requested):

earthkit/data/core/fieldlist.py
```diff
@@ -1164,6 +1164,7 @@ def _is_shared_grid(self):
                 )
         return False
 
+    @detect_out_filename
     def save(self, filename, append=False):
         r"""Write all the fields into a file.
```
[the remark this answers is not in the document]
EddyCMWF: docstring updated in: beb52aa

earthkit/data/readers/grib/reader.py
```diff
@@ -22,6 +22,7 @@ class GRIBReader(GribFieldListInOneFile, Reader):
     def __init__(self, source, path):
         Reader.__init__(self, source, path)
         GribFieldListInOneFile.__init__(self, path)
+        print("GribReader:", self.source_filename)
```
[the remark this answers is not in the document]
EddyCMWF: This I had already deleted, it was some code development debugging.

earthkit/data/readers/__init__.py
```diff
@@ -33,6 +33,11 @@ def __init__(self, source, path):
         self._source = weakref.ref(source)
         self.path = path
 
+        if hasattr(source, "source_filename"):
```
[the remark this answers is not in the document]
EddyCMWF: Went for the second option: acfc548

earthkit/data/sources/cds.py
```diff
@@ -125,7 +126,9 @@ def __init__(self, dataset, *args, **kwargs):
 
     def _retrieve(self, dataset, request):
         def retrieve(target, args):
-            self.client().retrieve(args[0], args[1], target)
```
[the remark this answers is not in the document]
EddyCMWF: source_filename implemented as a standard field in Cource class: 21e6041
Still thinking of an alternate name, maybe: `source_suggested_filename`

docs/examples/temp.nc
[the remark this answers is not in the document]
EddyCMWF: removed in: f71b895

earthkit/data/sources/cds.py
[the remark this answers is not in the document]
sandorkertesz: `source_suggested_filename` sounds better!

tests/sources/test_cds.py
```diff
@@ -120,6 +123,32 @@ def test_cds_grib_multi_var_date(date, expected_date):
...
+    with temp_directory() as tmpdir:
+        # Check file save to assigned filename
+        s.save(os.path.join(tmpdir, "test.grib"))
+        assert os.path.isfile(os.path.join(tmpdir, "test.grib"))
+
+        # Check file can be saved in current dir with detected filename:
+        here = os.curdir
+        os.chdir(tmpdir)
+        s.save()
+        assert os.path.isfile(os.path.basename(s.path))
```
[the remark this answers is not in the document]
EddyCMWF: As discussed, your suggestion is correct, but it demonstrated that the source_filename was not being resolved as a property correctly. This and the subsequent required fixed have been address in: 5aae0f8 
(debug print statements removed in: 015e0c2)

EddyCMWF:
Hi @sandorkertesz , tests are passing so hopefully good to go.
Thanks for the help!
Eddy

sandorkertesz (approved):

sandorkertesz:
Hi @EddyCMWF, thanks for the changes, it is now in a good shape. Please `squash and merge` this PR.

EddyCMWF:
> Hi @EddyCMWF, thanks for the changes, it is now in a good shape. Please `squash and merge` this PR.

Thanks Sandor, but I don't have permissions to do the squash and merge

sandorkertesz:
Hi @EddyCMWF, merged and the branch can be deleted.

PR #228, opened 2023-10-13
standardize requests to optimize caching

### LINKED ISSUE #208: Internally process the request kwargs to optimize caching

The following snippet triggers 2 requests from the CDS, although the only difference in the requests is the order of the variables requested.

```python
import earthkit.data
from earthkit.data import settings

settings.auto_save_settings = False
settings.set("cache-policy", "temporary")

request_kwargs = {
    "product_type": "reanalysis",
    "area": [50, -10, 40, 10],  # N,W,S,E
    "grid": [2, 2],
    "date": "2012-05-10",
    "time": "12:00",
}
for reverse in (True, False):
    ds = earthkit.data.from_source(
        "cds",
        "reanalysis-era5-single-levels",
        variable=sorted(["2t", "msl"], reverse=reverse),
        **request_kwargs
    )
```

Internally process the request kwargs to optimize caching.
For example, sort lists (all but area), squeeze single element lists (or do the opposite), change types when possible, ...

_No response_

### Additional context

_No response_

### Organisation

B-Open / CADS-EQC

malmans2:
Closes #208

### DIFF
--- a/earthkit/data/sources/cds.py
+++ b/earthkit/data/sources/cds.py
@@ -7,6 +7,8 @@
 # nor does it submit to any jurisdiction.
 #
 
+import collections
+
 import cdsapi
 import yaml
 
@@ -16,6 +18,12 @@ from earthkit.data.utils import tqdm
 
 from .file import FileSource
 from .prompt import APIKeyPrompt
+
+
+def ensure_iterable(obj):
+    if isinstance(obj, str) or not isinstance(obj, collections.abc.Iterable):
+        return [obj]
+    return obj
 
 
 class CDSAPIKeyPrompt(APIKeyPrompt):
@@ -103,7 +111,7 @@ class CdsRetriever(FileSource):
 
         return self.cache_file(
             retrieve,
-            (dataset, request),
+            (dataset, self.normalized_request(**request)),
             extension=EXTENSIONS.get(request.get("format"), ".cache"),
         )
 
@@ -121,6 +129,16 @@ class CdsRetriever(FileSource):
             r[split_on] = v
             result.append(r)
 
+        return result
+
+    @staticmethod
+    def normalized_request(**request):
+        result = {}
+        for k, v in sorted(request.items()):
+            v = ensure_iterable(v)
+            if k not in ("area", "grid"):
+                v = sorted(v)
+            result[k] = v[0] if len(v) == 1 else v
         return result
 
 
--- a/tests/sources/test_cds.py
+++ b/tests/sources/test_cds.py
@@ -67,6 +67,26 @@ def test_cds_grib_3():
 @pytest.mark.long_test
 @pytest.mark.download
 @pytest.mark.skipif(NO_CDS, reason="No access to CDS")
+def test_cds_normalized_request():
+    variables = ["2t", "msl"]
+    base_request = dict(
+        product_type="reanalysis",
+        area=[50, -50, 20, 50],
+        date="2012-12-12",
+        time="12:00",
+    )
+    s0 = from_source(
+        "cds", "reanalysis-era5-single-levels", variable=variables, **base_request
+    )
+    s1 = from_source(
+        "cds", "reanalysis-era5-single-levels", variable=variables[::-1], **base_request
+    )
+    assert s0.path == s1.path
+
+
+@pytest.mark.long_test
+@pytest.mark.download
+@pytest.mark.skipif(NO_CDS, reason="No access to CDS")
 def test_cds_netcdf():
     s = from_source(
         "cds",


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=228) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=228) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/earthkit-data?pullRequest=228) it.</sub>

malmans2:
Conflicts with develop fixed, this is ready for review.

The idea is to normalize the requests to optimize caching:
1. len 1 iterables are squeezed (probably I could have done the opposite as well, but I'm not sure if all request parameters allow lists + the request shown by the portal also uses strings rather than lists when a single element is selected)
2. sort len > 1 iterables, so the cache key generated is not affected by the order of the elements. There's a few parameters that should not be sorted. area and grid are the only ones I am aware of. The parameter that can not be sorted are currently hardcoded, I guess they could become a global variable

sandorkertesz:
Thank you for this PR. If it is just needed for the caching, one option is not to alter the request, just change the way the cache key is generated. E.g. in `cds.py`: 

```python
 def _retrieve(self, dataset, request):
        def retrieve(target, args):
            self.client().retrieve(args[0], args[1], target)

        # modify the request used for the cache key generation
        request = self._sort_request_for_caching(request)
        
        return self.cache_file(
            retrieve,
            (dataset, request),
            extension=EXTENSIONS.get(request.get("format"), ".cache"),
        )
```
In my opinion it would be a cleaner solution.

malmans2:
I thought about it, but I have 2 concerns:
1. This would trigger 4 requests rather than just 2:
```python
from_source("cds", ..., variable=["1", "3", "2", "4"], chunks={"variable": 2})
from_source("cds", ..., variable=["1", "2", "3", "4"], chunks={"variable": 2})
```
2. The public attribute `requests` of `CDSRetriever` does not show the request that has been actually submitted. I find it confusing from an user perspective.

I'll implement your suggestion if you still think that's better/cleaner.

sandorkertesz:
You are right.  To resolve point 2, I modified the code to separate the actual request from the one used for the cache key generation. Of course it still would not solve point 1 and I agree that you need to modify the actual request cover that case.

However, please note that `CDSRetriever` is not part of the public api and as such should not be used to retrieve data from the CDS. The public api is `from_source("cds", ...)`.

malmans2:
> Of course it still would not solve point 1 and I agree that you need to modify the actual request to cover that case.

How do you want to proceed? Should earthkit-data cover that case (leave this PR as it is) or not (implement [this](https://github.com/ecmwf/earthkit-data/pull/228#issuecomment-1772842700))? If the latter, in EQC we will standardize the requests before feeding them to earthkit-data (i.e., we don't need this PR).

> However, please note that CDSRetriever is not part of the public api and as such should not be used to retrieve data from the CDS. The public api is from_source("cds", ...).

If we don't use `CDSRetriever` directly, we can't associate the data downloaded with their requests (we need it to cache processed data as explained in point 3 [here](https://github.com/ecmwf/earthkit-data/pull/227#issuecomment-1764199585)). If EQC should not use `CDSRetriever.requests`, then we need to chunk the requests ourselves before feeding them to earthkit-data (i.e., we can't use the `split_on` functionality implemented in https://github.com/ecmwf/earthkit-data/pull/227)

sandorkertesz:
I still think `CDSRetriever` should not be used like that. This type of usage is not tested, can be refactored potentially breaking your code.

My suggestion is to pass an object to `from_source` that can perform the request preprocessing and able to return the actual requests sent to cds. E.g.

```python
   from earthkit.data import RequestPreprocessor
   rp = RequestPreprocessor()
   ds = from_source("cds",....., pre_processor=rp)
   requests = rp.request
```
```python
  class RequestPreprocessor:
     ....
     @property
     def request(self):
         ...
```

In this way the pre-processing could be reused for other retrieval types like "mars". And you would be able to subclass it and change the behaviour without having to modify earthkit-data.

An alternative solution is to somehow query the request form each retrieved file source object.

malmans2:
Being able to pass some sort of setup/teardown class that allows custom pre/post-processing would work for us. 

We can use it to pre-process the requests (e.g., the normalize decorators already implemented or the normalization in this PR) and post-process (e.g., global mean or time mean).

If you go that way and post-processed results are cached, then we don't even need to use our own caching. One important feature that we need is to be able to decide whether to post-process and cache each chunk separately (e.g., global mean & `split_on="date"`) or post-process requests after concatenating them (e.g., time mean & `split_on="date"`).

For reference, this is the sort of use EQC evaluators are currently doing:
```python
def global_mean(ds: xr.Dataset, weighted) -> xr.Dataset:
    if weighted:
        ds = ds.weighted(ds["area"])
    return ds.mean(("latitude", "longitude"))
    
def time_mean(ds: xr.Dataset) -> xr.Dataset:
    return ds.mean("time")


ds_global_weighted = download_and_transform(
    collection_id,
    requests,
    chunks={"year": 1, "month": 1},
    func=global_mean,
    kwargs={"weighted": False},
    transform_each_chunk=True,
)  # First, cache raw data. Then, cache the global mean of each chunk

ds_global_weighted = download_and_transform(
    collection_id,
    requests,
    chunks={"year": 1, "month": 1},
    func=global_mean,
    kwargs={"weighted": True},
    transform_each_chunk=True,
)  # Cache the global weighted mean of each chunk re-using cached raw data

ds_time = download_and_transform(
    collection_id,
    requests,
    chunks={"year": 1, "month": 1},
    func=time_mean,
    transform_each_chunk=False,
)  # Cache the time mean of all chunks concatenated re-using cached raw data
```


If you decide to implement https://github.com/ecmwf/earthkit-data/pull/228#issuecomment-1776781130, feel free to ping me if you'd like me to test it out.

malmans2:
Talking with @EddyCMWF and @JamesVarndell I realised that the reason why we need to be able to associate each file to the corresponding request was not clear. Here is a simpler example:
```python
from earthkit.data.sources.cds import CdsRetriever

import cacholote
import xarray as xr


@cacholote.cacheable
def _global_mean(collection_id: str, request: dict[str, Any]) -> xr.Dataset:
    ds = from_source("cds", collection_id, request).to_xarray()
    return ds.mean(("x", "y"))


def global_mean(collection_id: str, *args: dict[str, Any], **kwargs: Any) -> xr.Dataset:
    datasets = []
    for request in CDSRetriever(collection_id, *args, **kwargs):
        datasets.append(_global_mean(collection_id, request))
    return xr.merge(datasets)


ds_100yrs = global_mean("collection_id", ..., year=range(1900, 2000), split_on="year")
print(ds_global_mean.chunksizes)  # {"year": (1, ) * 100}

# Use cached data
ds_10yrs = global_mean("collection_id", ..., year=range(1990, 2000), split_on="year")
print(ds_global_mean.chunksizes)  # {"year": (1, ) * 10}
```

sandorkertesz:
Hi @malmans2, I am sorry I could not work on this PR for a while. I will resume reviewing next week.

Issue #177, opened 2023-09-07
Test geojson_bounding_box fails


sandorkertesz:
### What happened?

`test_geojson_bounding_box` fails with the following error in one of my conda envs on ATOS.

Versions:
geopandas                   0.13.2
pandas                        1.5.2

```
pytest -v -k test_geojson_bounding_box
==================================================================== test session starts =====================================================================
platform linux -- Python 3.8.12, pytest-7.2.0, pluggy-1.0.0 -- conda/envs/pyfdb/bin/python3.8
cachedir: .pytest_cache
rootdir: /git/earthkit-data, configfile: pytest.ini, testpaths: tests
plugins: forked-1.6.0, timeout-2.1.0
collected 718 items / 717 deselected / 1 selected                                                                                                            

tests/readers/test_geojson_reader.py::test_geojson_bounding_box FAILED

========================================================================== FAILURES ==========================================================================
_________________________________________________________________ test_geojson_bounding_box __________________________________________________________________

    def test_geojson_bounding_box():
        s = from_source("file", earthkit_file("tests/data/NUTS_RG_20M_2021_3035.geojson"))
>       bb = s.bounding_box()

tests/readers/test_geojson_reader.py:26: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
earthkit/data/readers/geojson.py:64: in bounding_box
    return self._to_geopandas_dataframe_wrapper(**kwargs).bounding_box()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <earthkit.data.wrappers.pandas.GeoPandasDataFrameWrapper object at 0x154abedb1eb0>

    def bounding_box(self):
        """
        For geopandas, get bounding box and convert to EK.BoundingBox type
        """
        return BoundingBox(
[...]

PR #239, opened 2023-10-18
handle bad installations of PROJ

EddyCMWF:
Current implementation results in failed tests if the underlying installation of PROJ is not compatible.
This change attaches a marker to the test such that it can be omitted by the user.
The environments used by the CI tests do not have this issue, so it is not necessary to omit the test for the GH actions.

Also included is the assignment of NaN values to the boundingbox such that the attributes exist, they are just unknown, this would allow users to set by hand if desired.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -7,4 +7,5 @@ markers =
     notebook: testing notebooks can be slow. But needs to be performed to ensure that the documention is tested.
     no_cache_init: a test where the cache is not initialised. Must be run with --forked.
     no_eccodes: a test which should pass when ecCodes is not installed
+    with_proj: a test which should only pass if PROJ is installed and compatible with your system
 testpaths = tests
--- a/earthkit/data/wrappers/pandas.py
+++ b/earthkit/data/wrappers/pandas.py
@@ -8,6 +8,8 @@
 
 import logging
 
+import numpy as np
+
 from earthkit.data.utils.bbox import BoundingBox
 from earthkit.data.wrappers import Wrapper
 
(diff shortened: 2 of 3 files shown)


pytest.ini
```diff
@@ -7,4 +7,5 @@ markers =
     notebook: testing notebooks can be slow. But needs to be performed to ensure that the documention is tested.
     no_cache_init: a test where the cache is not initialised. Must be run with --forked.
     no_eccodes: a test which should pass when ecCodes is not installed
+    with_proj: a test which should only pass if PROJ is installed and compatible with your system
```
sandorkertesz, line 10: I wonder if we can detect it automatically as we do for other modules in `testsing.py, e.g. for pyfdb.

sandorkertesz: I created the conda env on atos with the standard yaml above and got the error I reported in #177.

sandorkertesz: The ci tests are not running the `with_proj` test automatically. Please modify the`unit-tests` in `Makefile` to make it work.

EddyCMWF: I think the ci jobs are running the test, unless I am misunderstanding what you mean. For me they run fine locally with `make unit-tests` and the github actions see them and run them:

image

sandorkertesz: I am sorry you are right.

sandorkertesz:
This PR is ready to merge, but first the branch has to be updated since it is `out-of-date with the base branch`.

EddyCMWF:
> This PR is ready to merge, but first the branch has to be updated since it is `out-of-date with the base branch`.

done

PR #268, opened 2023-12-01
cfgrib does not need to be specified. 

EddyCMWF:
cfgrib engine automatically detects when it is needed, hene it is not neccesary to include in open_dataset kwargs.
Additionally current implementation prevents using different engine (which we are working on)


sandorkertesz:
@EddyCMWF, what is the "different engine (which we are working on)"?

PR #269, opened 2023-12-04
Feature/optional dependencies

### LINKED ISSUE #252: Make some dependencies optional
### What maintenance does this project need?

Requirements file needs to be revised for earthkit-data as it installs too many things. We could solve that through optional dependencies in the `toml` config, for instance having an option to install all dependencies, or just a subset, or just the basic ones.

### Organisation

ECMWF

sandorkertesz:
Fixes #252 

See docs at: https://earthkit-data--269.org.readthedocs.build/en/269/install.html

The optional Python dependencies are now controlled in `setup.cfg`. 

By default when installed from PyPI we get a minimal version:

image

Optional dependencies can be installed as:

image


sandorkertesz:
Some more questions:
- should we make "grib" or "netcdf" as an optional dependency?
- how should `earthkit` and `earthkit-maps` define `earthkit-data` as a dependency? Should they specify `earthkit-data[all]` ???
- how should the `earthkit-data` conda install behave?

iainrussell (changes requested):
Looks good to me! I see we mention eccodes twice as a dependency (generally, and for BUFR support). Is that because BUFR support might require a newer version? I'd say that since both instances of the eccodes dep currently require the same version, we could drop it as a separate dep for BUFR.

iainrussell:
Great, I think we're ready to merge then. Nice work!

iainrussell (approved):

sandorkertesz:
> Great, I think we're ready to merge then. Nice work!

Thank you! What do you think about the questions above?

iainrussell:
> Some more questions:
> 


Sorry, I did not notice those questions.

> * should we make "grib" or "netcdf" as an optional dependency?
I think for now they should be always there, as they are 'core' data types that we support, and the dependencies are easy to install everywhere.

> * how should `earthkit` and `earthkit-maps` define `earthkit-data` as a dependency? Should they specify `earthkit-data[all]` ???
Good question. Probably 'earthkit' should install all, and ek-maps should be more conservative, but this is a slightly different question from this specific PR. And it depends a bit on whether lacking the binary dependencies would cause an error during 'import earthkit.data' in the 'all' case.

> * how should the `earthkit-data` conda install behave?
Probably it should use 'all', but this will not work (yet) in practice because we do not have all the binaries available on conda-forge (e.g. odc, fdb). Once we have our own conda channel up and running, we will be able to do this. So, we should just put all the dependencies that are available on conda-forge as dependencies of the conda-forge package.

sandorkertesz:
> > * how should `earthkit` and `earthkit-maps` define `earthkit-data` as a dependency? Should they specify `earthkit-data[all]` ???
> 
> Good question. Probably 'earthkit' should install all, and ek-maps should be more conservative, but this is a slightly different question from this specific PR. And it depends a bit on whether lacking the binary dependencies would cause an error during 'import earthkit.data' in the 'all' case.

'import earthkit.data' should be fine even when the binary dependencies are missing. This PR is only addressing the Python dependencies.

PR #318, opened 2024-02-23
Add source to load sample data

sandorkertesz:
This PR adds the `from_sample` method to access data stored on the download server in folder `earthkit/samples`. The primary goal is to help users to quickly access data used for the examples in `earthkit-maps` and `earthkit-data`.

This is how it works:

```python
from earthkit.data import from_sample
ds = from_sample("storm_ophelia_wind_850.grib")
```

Alternatively we could introduce another source "sample" to implement this feature, e.g.:

```python
ds = from_source("sample", "storm_ophelia_wind_850.grib")
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -92,6 +92,7 @@ html_logo = "_static/earthkit-data.png"
 
 xref_links = {
     "cfgrib": ("cfgirb", "https://github.com/ecmwf/cfgrib"),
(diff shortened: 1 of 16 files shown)


iainrussell (commented):
I marginally prefer the from_source("sample", ...) option for this for consistency. It's a really handy function to have though!

sandorkertesz:
The feature is now implemented as a sample source:

image

PR #319, opened 2024-02-23
from_source should fail when try to load empty file

sandorkertesz:
With this PR `from_source` throws an exception when the input is an empty file. Previously, the result was a `TextReader`.

The consequence of this change is that when reading/downloading multiple files now it is enough to have an empty file out of many valid ones and the program fail if the exception is uncaught.


iainrussell (approved):
This behaviour seems pretty reasonable to me. If someone encounters a real use case where they don't want this, we can reconsider.

PR #340, opened 2024-03-09
Restructure examples

sandorkertesz:
This PR restructures and renames the examples to cope with the growing number of notebooks and offer better navigation. See: https://earthkit-data--340.org.readthedocs.build/en/340/examples/index.html


iainrussell (changes requested):
Hi Sandor, this is looking much better now!
I suggest the following refinements before merging:

- 'Sources' should be 'Data sources'
- 'Other inputs' is duplicated

sandorkertesz:
Hi @iainrussell, it is fixed now.

iainrussell (approved):
Looks great, thank you! It's also really nice that you managed to get the sidebar working properly!

PR #414, opened 2024-07-05
modify FDB env var checks

cosunae:
This PR tries to solve two issues: 
- FDB library can be set through two env variables, FDB_HOME or FDB5_HOME. earthkit-data requires here FDB_HOME but pyfdb requires FDB5_HOME: https://github.com/ecmwf/pyfdb/blob/b004bd58ed18c3d7e8bfc3e8081ae6c24fc4b739/pyfdb/pyfdb.py#L44
it would be better if all python ecosystem using FDB would require the same variable, therefore I propose here to use FDB5_HOME to be compatible with pyfdb.
- The config can be set through FDB5_CONFIG or FDB5_CONFIG_FILE.


sandorkertesz:
Hi @cosunae, thank you for this PR. 

I agree with you that the checking for FDB support should be more consistent. However, please consider these points:

1. pyfdb finds the fdb5 library by using `findlibs`:

    ```python
     libName = findlibs.find("fdb5")
    ```
  
     This call tries to find it by applying the algorithm described at: https://github.com/ecmwf/findlibs/tree/develop. It checks many   things e.g. env variables like FDB5_HOME, FDB5_DIR etc, but it also checks e.g. the conda env if these are not available. 

2. The checks in `sources/fdb.py` are meant to test the availability of the FDB config rather than that of the library, since if the library is not available pyfdb will raise an exception anyway. 

3. The checks you propose are too strict because with this I need both FDB5_HOME and FDB5_CONFIG/FDB5_CONFIG_FILE to be defined.  

    ```python
     if fdb5_home is None or (fdb5_conf is None and fdb5_config_file is None):
                raise ValueError(
    ```
    However, at the moment I am able to use pyfdb in my conda env (containing the FDB5 lib) having only FDB_HOME (!) defined.

cosunae:
Thanks @sandorkertesz 
let me reply to your comments: 
1. yes, my understanding of the find_libs algorithm is that it will mainly check for FDB5_XXX instead of FDB_XXX, therefore the reason for this proposal. Is this assumption right ? 
2. understood, thanks for the clarification. 
3. I see, yes, if FDB5_HOME is enough we should not make it stricter. I changed it to have any of the variables defined, can you check now ?

sandorkertesz:
>     3. I see, yes, if FDB5_HOME is enough we should not make it stricter. I changed it to have any of the variables defined, can you check now ?
The new check seems to be fine now! However, internally we do use FDB_HOME, so I suggest you add it back and check its existence together with the other 3. Then this PR can be merged.

PR #428, opened 2024-08-05
Add GRIB field, handle and metadata cache

sandorkertesz:


docs/guide/misc/grib_memory.rst
```diff
@@ -0,0 +1,91 @@
...
+++++++++++++++++++++++++++++
+
+When ``store-grib-fields-in-memory`` is ``True`` (this is the default) once a field is created it will be stored in the fieldlist. Otherwise the field will be created on demand and deleted when going out of scope.
+
+The actual memory used by a field depends on whether it stores the GRIB handle of the related GRIB message. This is controlled by the :ref:`grib-handle-cache-size <grib-handle-cache-size>` settings:
+
+ - When ``grib-handle-cache-size > 0`` the field objects themselves are lightweight and only store the GRIB handle cache index, while the actual GRIB handles are stored in the cache, which is attached to the fieldlist.
+ - When ``grib-handle-cache-size == 0`` the behaviour depends on ``store-grib-fields-in-memory``:
+
+    - when ``store-grib-fields-in-memory`` is ``True`` the fields does not own their GRIB handle but for each call to data and metadata access a new GRIB handle is created and released once the access finished. This can be useful when the fields have to be kept in memory but the memory usage should be low.
```
iainrussell, line 74: Is this the right way round? This sounds like what I'd expect if the flag was False

docs/guide/misc/grib_memory.rst
```diff
@@ -0,0 +1,91 @@
...
+    - when ``store-grib-fields-in-memory`` is ``False`` the fields are created on demand and will store their own GRIB handle in memory until they gets deleted (when going out of scope).
+
+
+.. _grib-handle-cache-size:
+
+grib-handle-cache-size
+++++++++++++++++++++++++++++
+
+When ``grib-handle-cache-size`` is set to a positive value (default is 1) an in-memory GRIB handle cache is attached to the fieldlist. The cache size determines how many GRIB handles are stored in memory. This is an LRU cache so when it is full the least recently used GRIB handle is removed and a new GRIB message is loaded from disk and added to the cache.
```
iainrussell, line 83: Is this specified in bytes, or the number of handles? The wording is not 100% clear

sandorkertesz: Number of handles. I improved the explanation.

docs/guide/misc/grib_memory.rst
sandorkertesz: Following up our discussion the setting were completely refactored.

sandorkertesz:
@iainrussell I wonder if the `grib-handle-cache-size` is the right name to use. In the settings:

https://earthkit-data.readthedocs.io/en/latest/guide/settings.html

we already have `maximum‑cache‑size`. So the settings should ideally be:  `maximum-grib-handle-cache-size`?

PR #493, opened 2024-10-21
Alter field metadata

sandorkertesz:
This PR adds the `copy()` method to fields. The result is a new field containing:

- reference to the values/values accessor in the original field
- a wrapped metadata object updated with the kwargs

The original field is kept unaltered.

E.g. if `f` is a field with "level" = 500:

```python
>>> f_new = f.copy(level=700)
>>> f.metadata("level")
>>> f_new.metadata("level")
500
700
```

The kwargs can contain callables with the signature described below:

```python

def _func(field, key, original_metadata):
    return original_metadata[key] + 100

f_new = f.copy(level=_func)
print(f.metadata("level"))
print(f_new.metadata("level"))
```
```
500
600
```

### DIFF
--- a/src/earthkit/data/core/fieldlist.py
+++ b/src/earthkit/data/core/fieldlist.py
@@ -79,12 +79,6 @@ class FieldListIndices:
 class Field(Base):
     r"""Represent a Field."""
 
-    def __init__(
-        self,
-        metadata=None,
-    ):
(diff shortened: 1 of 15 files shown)


b8raoult:
`original_metadata` is a dict(), not a `Metadata`

b8raoult (approved):

Issue #120, opened 2023-06-16
How to support geotiff?


sandorkertesz:
This issue was created to discuss the possible support for geotiff  in earthkit-data.

milanavuckovic:
Thanks for opening this issue, Sandor.
Support for geotiff data would be much appreciated in CLIMAAX project.

sandorkertesz:
I tried to use https://github.com/KipCrossing/geotiff to read geotiff, a `noGdal` project, but I immediately hit a limitation with the geotiffs from the CLIMAAX Land use data:

```
  255     return temp_crs_code
    256 elif temp_crs_code == 32767:
--> 257     raise UserDefinedGeoKeyError(crs_error_msg)
    258 else:
    259     raise GeographicTypeGeoKeyError()

UserDefinedGeoKeyError: User-defined GeoKeys are not yet supported.
```

The githiub page actually mention it as a core feature yet to be implemented:
"read a user defined CRS 32767 from tiff file"

So I am not sure we can use it in earthkit-data.

corentincarton:
Tagging @chpolste, who will investigate this for the CLIMAAX project.

corentincarton:
From @colonesej, this link could be interesting: https://kipling.medium.com/nogdal-e5b60b114a1c

sandorkertesz:
Fixed by #503

milanavuckovic:
[celebrate]     Milana Vuckovic reacted to your message:
________________________________

PR #503, opened 2024-11-01
Reading GeoTIFFs with rioxarray/rasterio

chpolste:
A basic reader for GeoTIFF files using rioxarray under the hood. To address #120.

- Implements `GeoTIFFReader`, `GeoTIFFFieldList`, `GeoTIFFField`, `GeoTIFFMetadata` and `GeoTIFFGeography`.
- The unit test comes with a small multi-band test file.
- New `geotiff` optional dependencies in `pyproject.toml`.

There is not much in terms of metadata extraction other than the projection information yet, but many GeoTIFFs don't provide much metadata in the first place, unfortunately.

I have configured the reader to apply masking and scaling to the values by default, even though rioxarray doesn't.

Small things adjusted along the way:

- Turned `shape` of the `core.geography.Geography` abstract class into a method (was a property), since that is how it seems to be used in practice.
- Added a `TransverseMercator` projection, which seems to be common for GeoTIFF data.
- Changed `utils.projections.cf.CF_PARAMS_TO_GLOBE_KWARGS["inverse_flattening"]` to `"inverse_flattening"` (was `"flattening"`).


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=503) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf/earthkit-data?pullRequest=503) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf/earthkit-data?pullRequest=503) it.</sub>

src/earthkit/data/readers/geotiff.py
```diff
@@ -0,0 +1,219 @@
...
+        self._not_implemented()
+
+    def bounding_box(self):
+        left, bottom, right, top = self._ds.rio.transform_bounds("EPSG:4326")
+        return BoundingBox(north=top, west=left, south=bottom, east=right)
+
+    def gridspec(self):
+        self._not_implemented()
+
+    def resolution(self):
```
sandorkertesz, line 77: This is a tricky method. For GRIB, returns a single value, and asserts if the resolution is different in x and y directions.

src/earthkit/data/readers/geotiff.py
```diff
@@ -0,0 +1,219 @@
...
+from functools import cached_property
+
+import numpy as np
+
+from ..core.fieldlist import Field
+from ..core.fieldlist import FieldList
+from ..core.geography import Geography
+from ..core.metadata import RawMetadata
+from ..utils.bbox import BoundingBox
+from ..utils.projections import Projection
```
sandorkertesz, line 20: Projection should be a local import in the methods that use it. With this we can delay importing e.g. catropy until it is actually used.

sandorkertesz:
@chpolste , thank you very much for this development.  I tested it and works nicely.  I made a few comments, once these and the branch conflict are sorted out the PR can be merged. 

As for the documentation, what notebook example do you think should be added? What should it be highlight?

tests/readers/test_geotiff_reader.py
```diff
@@ -0,0 +1,38 @@
...
+import pytest
+
+from earthkit.data import from_source
+from earthkit.data.readers.geotiff import GeoTIFFField
+from earthkit.data.testing import NO_RIOXARRAY
+from earthkit.data.testing import earthkit_test_data_file
+from earthkit.data.utils.projections import TransverseMercator
+
+
+@pytest.mark.skipif(NO_RIOXARRAY, "rioxarray not available")
```
sandorkertesz, line 21: I think this should be: 

```python
@pytest.mark.skipif(NO_RIOXARRAY, reason="rioxarray not available")
```

PR #529, opened 2024-11-18
Use eccodes.codes_set_key_vals in output.py

b8raoult:


sandorkertesz:
I fixed the failing tests. It revealed that we cannot set the `missing_value` to `np.finfo(np.float32).max`.

See the newly added test: test_grib_output_missing_value

PR #539, opened 2024-11-21
Update polytope example docs with latest code and visualisations

awarde96:


docs/examples/polytope.ipynb
```diff
@@ -54,146 +59,800 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c232c51b",
+   "metadata": {},
+   "source": [
+    "The collection being accessed is `ecmwf-mars`. The endpoint being accessed is `polytope.ecmwf.int`.\n",
+    "\n",
+    "To view the downloaded coveragejson you can run teh following cell."
```
sandorkertesz, line 103: should be the

docs/examples/polytope.ipynb
```diff
@@ -54,146 +59,800 @@
...
-    "import earthkit.data\n",
+    "ds.to_xarray()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f58a63b8",
+   "metadata": {},
+   "source": [
+    "## Visualise \n",
```
sandorkertesz, line 706: This should be a chapter inside "Time series" but "Visualise"  is now on the same level.

docs/examples/polytope.ipynb
```diff
@@ -54,146 +59,800 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "32cb0e10-7545-4758-b2f0-99984f01d71f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ds._json()"
```
sandorkertesz, line 113: _json should not be advertised. If we need to expose it a public method should added in a separate PR.

jameshawkes: does it make sense to expose it? I would think so, but not sure if this needs to be consistent with other data sources?

sandorkertesz: Not sure what the method name should be and what it should return. A dict?

sandorkertesz:
If we want to publish it without a release this should be branched off from "main" and then  merged into "main".

sandorkertesz:
What happened with the original content of this example. Is it outdated now?

jameshawkes:
You are right, maybe on this page we should just:
1. update the link to polytope readthedocs
2. add a sentence at the top saying that features can also be extracted, and link to the covjson examples

Sorry for the backtrack, @awarde96 could you do that?

sandorkertesz:
The polytope links should also be updated in `sources.rst`.

PR #540, opened 2024-11-22
Add feature extraction example notebook

awarde96:


sandorkertesz:
@awarde96, thank you for adding this new example. There are these minor issues to sort out before it can be merged:

- Please, add the new example to `docs/examples/index.rst`
- You also need to add it to the `SKIP` list in `tests/documentation/test_notebooks.py`

sandorkertesz:
There is a slight issue with the example. It now appears as 4 separate items in the CoverageJSON block:

image

The reason for this is that the chapter titles have the same size as the main title (##). I suggest you use ####  or ### for the chapter titles (this is what the other notebooks use). I also suggest you put this example into the Date sources block next to the polytope.ipynb example. Of course it can stay in CoverageJSON if you prefer that location but then the title should start with `CoverageJSON:`.

awarde96:
> There is a slight issue with the example. It now appears as 4 separate items in the CoverageJSON block:
> 
> image
> 
> The reason for this is that the chapter titles have the same size as the main title (##). I suggest you use #### or ### for the chapter titles (this is what the other notebooks use). I also suggest you put this example into the Data sources block next to the polytope.ipynb example. Of course it can stay in CoverageJSON if you prefer that location but then the title should start with `CoverageJSON:`.

I have moved it to the sources index, and hopefully the chapter titles are fixed

PR #565, opened 2024-12-11
Enable using environment variables for settings

### LINKED ISSUE #561: Enable controlling settings via environment variables

Currently, the settings can only be controlled via the `~/.earthkit-data/settings.yaml` file or from Python using the `earthkit.data.settings` interface. Additionally, it would be useful (e.g. in ecflow)  if we were able to control the settings via env variables too.

It is not clear whether we should add an env variable for each settings or only for a subset of these. The proposal here is to  only add env variables for the settings controlling the cache and the temporary directory:

- cache-policy
- temporary-cache-directory-root
- temporary-directory-root
- maximum-cache-size
- maximum-cache-disk-usage

The env variables should be all upper-case prefixed with EARTHKIT_DATA. E.g.:

cache-policy -> EARTHKIT_DATA_CACHE_POLICY

_No response_

_No response_

### Additional context

_No response_

### Organisation

ECMWF

sandorkertesz:
Fixes #561

### DIFF
--- a/docs/examples/index.rst
+++ b/docs/examples/index.rst
@@ -172,6 +172,7 @@ Miscellaneous
     :glob:
 
     settings.ipynb
+    settings_env_vars.ipynb
     cache.ipynb
     projection.ipynb
     metadata.ipynb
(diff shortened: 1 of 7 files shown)


iainrussell:
Yes, we agreed verbally that the env vars should simply always take precedence over the 'normal' settings. It's easier to understand.

PR #596, opened 2025-01-23
Writer api

sandorkertesz:
This PR

Adds a new writer API:
- Target 
- Encoder
- to_target()
- create_target()
- create_encoder()

https://earthkit-data--596.org.readthedocs.build/en/596/guide/targets.html
https://earthkit-data--596.org.readthedocs.build/en/596/guide/encoders.html

Deprecates:
 - write() and save() method
 - new_grib_output(), new_grib_coder()
 - readers.grib.output.GribOutput, readers.grib.output.GribCoder

Removes:
 - writers.GribWriter

Some counter-intuitive features
--------------------------------------

When we call `to_target()` it does the following:
- creates a target
- writes the data into the target
- calls close() on the target

This  has implications in the "file" target. Currently, when we call `FileTarget.close()` it behaves differently whether the target was created with a file path or a file pointer:

- if the FileTarget was created with a path `close()` **closes** the associated file pointer
- if the FileTarget was created with a file pointer `close()` **does not close** the associated file pointer

Examples (assume ds, ds1 and ds2 are fieldlists)

```python
# this opens the target file and closes it 
ds.to_target("file", "my.grib") 

fp = open("my_file", "wb")
ds1.to_target("file", fp)  # does not close fp
ds2.to_target("file", fp) # does not close fp
# fp is still open at this point, we need to close it explicitly
close(fp)

fp = open("my_file", "wb")
with create_target("file", fp) as t:
    for d in ds:
        t.write(data=d)

   # this will not close fp
   t.close()

# when t goes out of scope the context manager calls 
# close() which does not close fp
# fp is still open at this point, we need to close it explicitly
close(fp)
```

If we closed the user provided  file pointer when `close()` is called this code would not work:

```python
fp = open("my_file", "wb")
ds1.to_target("file", fp)  # close fp
ds2.to_target("file", fp) # ERROR: fp closed, cannot write to it
```


src/earthkit/data/encoders/netcdf.py
```diff
@@ -0,0 +1,74 @@
...
+class NetCDFEncodedData(EncodedData):
+    def __init__(self, ds):
+        self.ds = ds
+
+    def to_bytes(self):
+        return self.ds.to_netcdf(None)
+
+    def to_file(self, f):
+        if hasattr(self.ds, "earthkit"):
+            f.write(self.ds.earthkit.to_netcdf(None))
```
chpolste, line 27: Calling to_netcdf without a path restricts the data format to NetCDF3, btw:

> If no path is provided, this function returns the resulting netCDF file as bytes; in this case, we need to use scipy, which does not support netCDF version 4 (the default format becomes NETCDF3_64BIT).

https://docs.xarray.dev/en/stable/generated/xarray.Dataset.to_netcdf.html

sandorkertesz: @chpolste, thanks for pointing this out. Do you know if there is a way generate netCDF version 4 as bytes from xarray?

PR #675, opened 2025-04-09
Feature/zarr

Oisin-M:
Implements a simple `zarr` source for reading zarr files.


Oisin-M:
Tests are failing because zarr version 3 requires Python version 3.11 onward

sandorkertesz:
Hi @Oisin-M, thank you very much for this development. I had a quick look at the code and it seems to me that the "zarr" source is not really needed.

sandorkertesz (approved):
Thank you for adding this new source! Once it is in sync with the latest develop it can be merged.

sandorkertesz:
@oisin, unfortunately we cannot afford disabling all the tests due to this dependency. So I suggest you remove the zarr dependency from `environment-unit-tests.yml` and from "all" in `pyproject.toml`. It will be re-enabled once the CI python version is updated.

Oisin-M:
Hi @sandorkertesz, thanks for taking a look. I have now made zarr optional and also made the zarr test automatically skip if it can't find the zarr package, so tests are all passing

sandorkertesz:
@Oisin-M , I have a question about the source name. What was the reason of choosing "xarray_zarr" and not simply "zarr"? E.g. the "opendap" source is also using xarray to open the data but it does not have "xarray" in the name.

Oisin-M:
Hi @sandorkertesz, this was to leave room for having a reader based on the zarr API directly

Oisin-M:
Hi @sandorkertesz, I'd be happy to move towards that as long as we return objects with the same methods for both readers

sandorkertesz:
Well that is the catch. Not sure what `from_source("zarr", ..., reader="zarr")` would return. But I still think that the first argument in `from_source()` should describe the source and not the way we read it.

Issue #684, opened 2025-04-28
Add flatten argument to to_numpy in xarray wrappers


JamesVarndell:
### Is your feature request related to a problem? Please describe.

Many earthkit-data objects implement a `to_numpy()` method that accepts a `flatten` keyword argument. In other earthkit libraries, like earthkit-plots, we sometimes pass the `flatten` kwarg when calling `to_numpy`, depending on the use case.

However, the xarray Dataset and DataArray wrappers within earthkit-data do not support a `flatten` argument. This inconsistency can lead to errors when flatten is passed to their `to_numpy` methods. (See ecmwf/earthkit-plots#67).

### Describe the solution you'd like

- Update the xarray Dataset and DataArray wrappers in earthkit-data to accept a `flatten` keyword argument in their `to_numpy()` methods.
- If `flatten=True`, reshape the resulting NumPy array to be 1-dimensional after converting with `to_numpy()`.
- If `flatten=False` (default), maintain the current behavior and return the regular array.

This will ensure that all `to_numpy()` methods across earthkit objects have a consistent API and behavior.

### Describe alternatives you've considered

One alternative considered was handling this issue directly within earthkit-plots by wrapping calls to `to_numpy()` in a try-except block. If an error occurred due to an unexpected `flatten` argument, earthkit-plots could retry the call without the flatten parameter.

However, this approach felt suboptimal for several reasons:
- It would hide inconsistencies between earthkit-data object types rather than addressing the root cause.
- It would make the code more fragile and harder to maintain, as future changes to `to_numpy()` behavior across different objects could introduce subtle bugs.
- It would introduce unnecessary complexity into earthkit-plots, which should ideally be agnostic to these internal differences.

### Additional context

ecmwf/earthkit-plots#67

### Organisation

ECMWF

sandorkertesz:
Fixed by #685

PR #685, opened 2025-04-28
Adds flatten argument to to_numpy methods

JamesVarndell:
Adds the `flatten` argument to xarray DataArray and Dataset wrapper classes:
- If `flatten=True`, reshape the resulting NumPy array to be 1-dimensional after converting with `to_numpy()`.
- If `flatten=False` (default), maintain the current behavior and return the regular array.

See #684 for more information.


sandorkertesz:
@JamesVarndell, thank you very much for this development. Please can I ask you to add a test to `tests/wrappers/test_xarray.py`?

PR #688, opened 2025-05-02
improve/fix xr splitter

pawel-wolff:
xr_engine splitter improved/fixed in case "split_dims" contains multiple keys

If to_xarray converter gets "split_dims" keyword, a slight change of behaviour is introduced: it returns a tuple with:
- a list of xarray Datasets
- a corresponding list of dictionaries with values of the spitting keys.
Moreover, the two above are always lists, even if they are of length 1 (for consistency of the behaviour).

TBD if we really want to introduce this new behaviour since the values of the splitting keys can be obtained via using "ensure_dims" keyword of to_xarray.

A test and a corresponding mid-size test dataset was added.


sandorkertesz:
@pawel-wolff, thank you very much for this fix and improvement in the Xarray engine. I fully agree with your decision to change the return type when ``split_dims=True``.

PR #723, opened 2025-06-13
Rename xarray-zarr source and add docs

sandorkertesz:
This PR renames the `xarray-zarr` source to `zarr`. Also adds documentation.

The reason behind the renaming:

The name  "xarray-zarr" reflects the fact that the Zarr store is read by Xarray internally. The suggestion is to use a more generic name "zarr". In the future if another Zarr reader is added e.g. based on the Zarr API we could handle it with an extra kwarg in `from_source()`. E.g. 

```python
from_source("zarr", ..., reader="xarray")
from_source("zarr", ..., reader="zarr")
```

Also consider that none of the existing source names contains/refers to the reader.


src/earthkit/data/readers/directory.py
```diff
@@ -69,7 +69,7 @@ def mutate_source(self):
         ):
             if self.stream:
                 raise ValueError("Cannot stream zarr directories")
-            return from_source("xarray-zarr", self.path)
+            return from_source("xarray", self.path)
```
Oisin-M, line 72: Should be `from_source("zarr", self.path)`

PR #802, opened 2025-09-15
Fix handling empty slices of a tensor

pawel-wolff:
### Description

This PR fixes a problem in handling empty slices of Xarray objects whenever the underlying data in managed by a field list through a `FieldListTensor` object (e.g. when a GRIB file is converted to Xarray via `.to_xarray(lazy_load=True, ...)` method). Consider the following code:

```python3
from earthkit.data.testing import earthkit_remote_test_data_file

ds_ek = from_source("url", earthkit_remote_test_data_file("xr_engine/level/pl.grib"))
ds = ds_ek.to_xarray(lazy_load=True)

ds2 = ds.sel(level=slice(350, 360))  # in ds2, the level-dim has size 0
ds2.load()  # crashes

ds3 = ds.load().sel(level=slice(350, 360))   # this works, because the full ds is loaded into memory 
                                             # before extracting an empty slice
```

The problem with `ds2.load()` comes from the fact that the underlying field list is empty (i.e. `None`, see https://github.com/ecmwf/earthkit-data/blob/195942d2badd39fea1569c810af8f2686d1cfa9a/src/earthkit/data/core/fieldlist.py#L1108). 

This PR fixes this issue by properly handling the case of an empty field list, see: https://github.com/ecmwf/earthkit-data/blob/dbc9810475a158469db6da55c1036271b23c0ecc/src/earthkit/data/indexing/tensor.py#L433

A related code was refactored slightly to remove some code duplication between `to_numpy`, `to_array` methods within `FieldListTensor` and `FieldListSparseTensor` classes.

A relevant test was added.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/indexing/tensor.py
+++ b/src/earthkit/data/indexing/tensor.py
@@ -15,6 +15,7 @@ from abc import abstractmethod
 
 import numpy as np
 from earthkit.utils.array import array_namespace
+from earthkit.utils.array import get_backend
 
 from earthkit.data.core.index import Selection
 from earthkit.data.core.index import normalize_selection
@@ -394,16 +395,16 @@ class FieldListTensor(TensorCore):
         self._full_shape = None
         self.flatten_values = None
 
-    def _prepare_tensor_data(self, source_as_array_method, index=None, array_backend=None, **kwargs):
+    def _prepare_tensor_data(self, source_to_array_func, dtype=None, array_backend=None, index=None):
         if index is not None:
             if all(i == slice(None, None, None) for i in index):
                 index = None
 
         if index is None:
-            arr = source_as_array_method(array_backend=array_backend, **kwargs)
+            arr = source_to_array_func()
             current_field_shape = self.field_shape
         else:
-            arr = source_as_array_method(index=index, array_backend=array_backend, **kwargs)
+            arr = source_to_array_func(index=index)
             if arr is not None:
                 current_field_shape = tuple(arr.shape[1:])
             else:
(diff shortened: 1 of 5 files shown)


src/earthkit/data/indexing/tensor.py
```diff
@@ -394,33 +394,57 @@ def clear(self):
         self._full_shape = None
         self.flatten_values = None
 
-    @flatten_arg
-    def to_numpy(self, index=None, **kwargs):
+    def _prepare_tensor_data(self, method, index=None, **kwargs):
+        if method == "to_array":
```
sandorkertesz, line 398: The method itself can be passed as  `method` to avoid ifs and strings.

src/earthkit/data/indexing/tensor.py
```diff
@@ -394,33 +394,57 @@ def clear(self):
...
+                    len(range(n)[_slice])
+                    for n, _slice in zip(self.field_shape, index)
+                    if not isinstance(_slice, int)
+                    # `_slice` can be either an `int` or a `slice`; if `int`, ignore it!
+                )
+        return arr, current_field_shape
+
+    def _to_array(self, method, index=None, **kwargs):
+        arr, current_field_shape = self._prepare_tensor_data(method, index=index, **kwargs)
+        if arr is None:
```
sandorkertesz, line 435: This should be array format agnostic. The array_namespace is ideally in the kwargs.

tests/xr_engine/test_xr_tensor.py
```diff
@@ -0,0 +1,107 @@
...
+                {"step": "0h", "values": slice(0)},
+                {"step": slice("1h", "5h"), "values": slice(0)},
+                {"step": "0h", "values": [1, 2, 10]},
+                {"step": slice("1h", "5h"), "values": [1, 2, 10]},
+            ],
+            [(1, 1122), (1, 0), (0,), (0, 0), (3,), (0, 3)],
+        ),
+    ],
+)
+def test_xr_tensor_empty_slices(allow_holes, lazy_load, file, variables, sel_dicts, shapes):
```
sandorkertesz, line 95: This is a great test. But I would remove "tensor" from the name since you do not explicitly use tensor here. I would also rename the file itself e.g. "test_xr_slice.py" since it is about slicing.

src/earthkit/data/indexing/tensor.py
```diff
@@ -394,33 +394,57 @@ def clear(self):
...
+            import earthkit.utils.array.namespace.numpy as xp
+
+            arr = xp.empty(shape=0, dtype="f8")
+        return arr.reshape(self.user_shape + current_field_shape)
```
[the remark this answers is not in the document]
pawel-wolff: Do you mean using the top-level function `xp.reshape(arr, new_shape)` for an array namespace `xp` rather than `arr.reshape(new_shape)` ?

src/earthkit/data/indexing/tensor.py
pawel-wolff: Done in this commit: 0c5a742b944d34be780b1ce58d582cd63e99aab7

src/earthkit/data/indexing/tensor.py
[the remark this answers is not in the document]
sandorkertesz: Yes, exactly. We try to use all these methods on the array namespace (when it supports the methods). Both `empty` and `reshape` are part of the array api.

PR #902, opened 2026-02-18
Progress with xr-engine

pawel-wolff:
### Description

This PR provides many adaptations to xr-engine in order to make it compatible with the new field implementation. Also, a few bugs in `earthkit.data.core` are fixed. Here is a highlight of the main changes:

* All time and level dim modes work
* Dimensions, attributes and related xr-engine kwargs work (except for namespace-related - to be implemented)
* In an xarray object, the dimension and attributes names have the `"<component>."` prefix stripped
* Hard-coded dimension keys in `xr_engine.dim` adapted to the new field.
* Profiles updated (as for now, only legacy profiles provided: `"default"` (`None`), `"grib"`, `"mars"`)
* Selection (`.sel()`) on a field list fixed: a bug in `core.index.Selection.InList.__call__` fixed
* The PR #901 merged

Many xr-engine tests adapted. Among xr-engine tests 463 are passing / 331 are failing

### Contributor Declaration


src/earthkit/data/core/order.py
```diff
@@ -69,6 +69,16 @@ def substitute(self, name, joiner, **kwargs):
             return joiner.join(lst)
         return joiner.format_name(name, **kwargs)
 
+    def components(self, name):
```
sandorkertesz, line 72: I do not think this is needed.

pawel-wolff: It is actually used in `DimHandler._init_remapping_dims`

https://github.com/ecmwf/earthkit-data/blob/f69e2bea78695b661c748c768c9ccff0c674394b/src/earthkit/data/xr_engine/dim.py#L739

where the result of the call `remapping.components()` is passed to the `RemappingDim` init

https://github.com/ecmwf/earthkit-data/blob/f69e2bea78695b661c748c768c9ccff0c674394b/src/earthkit/data/xr_engine/dim.py#L386

I've checked, and this piece of code is indeed used whenever `to_xarray` has a `remapping` defined.

Do you have another solution in mind?

sandorkertesz: Thanks! That is fine then. This can be reviewed later. I will re-add tests related to this method.

PR #932, opened 2026-03-18
Perf/better tensor indexing

pawel-wolff:
### Description

This PR addresses a performance issue which arises in certain xarray computational workflows involving earthkit lazy loaded GRIB data. 

For example, consider a 3-year time-series temperature data with 6h resolution (~4k fields, stored in GRIB). Calculating climatological daily mean for a given day of year requires taking an average of 12 fields scattered along the whole time axis (4 fields per day x 3 years). The existing indexing method of an earthkit-data tensor allows for `BASIC` indexing method only: each axis can be indexed with an int or a slice (see https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html#indexing-examples):
https://github.com/ecmwf/earthkit-data/blob/11e7ed5ad1ae94e48b5e5926f821e98baba0373b/src/earthkit/data/xr_engine/builder.py#L238
In turns out, that when the GRIB data is opened with earthkit and converted into an xarray object with `.to_xarray()` method using (the default) lazy-loading, then the xarray calculation
```
ds.groupby("valid_time.dayofyear").mean()
```
triggers the following access pattern to the underlying GRIB fields: for each day of a year (`d`), a slice of the 3-year time-series of fields spanning from `year0 + day d` to `year2 + day d` (which covers ~2/3 of the time-series) is accessed and converted to (a numpy) array. In total, the above calculation of the climatological daily mean triggers ~365 x (2/3) x 4k accesses to the GRIB fields, or, in other terms, each of the ~4k fields is accessed ~ 365 x (2/3) times. Of course, an optimal execution of such a workflow should access each field just once.

The solution proposed here is to extend the earthkit tensor indexing to `OUTER` method, which, besides a single int or a slice, allows a *list of int's* as an indexer. In the scenario of the same computation as above (climatology daily mean), it allows xarray to use much more efficient access pattern to the underlying GRIB fields: for each day of year, only 12 fields that are actually needed are being accessed. It means that each field is accessed once. In this way, the execution time of the example computation drops from 3-5 minutes (depending on the platform) to ~1 sec.

A performance test based on the above example computation workflow and its variants has been added. A generous timeout  constraint (~3x larger than needed for a MacBook laptop) is imposed.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/indexing/tensor.py
+++ b/src/earthkit/data/indexing/tensor.py
@@ -236,7 +236,8 @@ class TensorCore(metaclass=ABCMeta):
                     if not isinstance(v, (list, tuple, np.ndarray)):
                         v = (v,)
                     r[k] = v
-                elif isinstance(idx, list):
+                else:
+                    # idx must be an iterable (list, numpy array, etc.)
                     r[k] = tuple([self._user_coords[k][i] for i in idx])
             else:
                 r[k] = self._user_coords[k]
@@ -400,7 +401,7 @@ class FieldListTensor(TensorCore):
 
     def _prepare_tensor_data(self, source_to_array_func, index=None):
         if index is not None:
-            if all(i == slice(None, None, None) for i in index):
+            if all(isinstance(i, slice) and i == slice(None, None, None) for i in index):
                 index = None
 
         if index is None:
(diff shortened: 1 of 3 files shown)


sandorkertesz:
@pawel-wolff Many thanks for this massive improvement.

PR #933, opened 2026-03-18
Source to access data from Zenodo based on record ID or DOI

chpolste:
### Description

- The Zenodo source translates record IDs and DOIs from the Zenodo namespace into URLs for individual files, then delegates to a `MultiUrl` source for accessing the files.
- The source offers file name matching based on Python's `fnmatch`, so users can select, e.g., all netcdf files of a record conveniently. Files can also be selected explicitly by providing a list of files.
- Adds `requests` as an explicit dependency. As far as I can tell, it was already pulled in by the `multiurl` dependency previously.

[docs preview](https://earthkit-data--933.org.readthedocs.build/en/933/concepts/inputs/from_source.html#data-sources-zenodo)

### Usage

```python
# Access with record id, select all files by default
ekd.from_source("zenodo", 10458186)

# Access with DOI, select only files with grib extension
ekd.from_source("zenodo", "10.5281/zenodo.10458186", only="*.grib")

# Concept DOI is resolved to latest version, DOI accepted as URL, select specific files
ekd.from_source("zenodo", "https://doi.org/10.5281/zenodo.10458185", only=["FWI_T0.0_P0.8_EU_1981_2010.grib", ...])
```

### Tests

The current test suite only checks that the Zenodo source produces correct URLs. It mocks the `requests` module and produces responses as expected from the Zenodo API to avoid spamming API requests to the actual Zenodo API.

### Questions

- Is `only` a good name for the file selection argument?
- Should the test suite have a test that verifies against the actual Zenodo API? If so, I could create a dedicated Zenodo test data record with a few example files.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/sources/zenodo.py
+++ b/src/earthkit/data/sources/zenodo.py
@@ -1,0 +1,65 @@
+# (C) Copyright 2026- ECMWF and individual contributors.
+
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation nor
+# does it submit to any jurisdiction.
+
+import fnmatch
+import re
+
+import requests
+
+from earthkit.data.sources import Source
+from earthkit.data.sources import from_source_internal
+
+_DOI_PATTERN = re.compile(r"^(?:doi:\s*)?(10\.5281/zenodo\.\d+)$", flags=re.IGNORECASE)
+_URL_PATTERN = re.compile(r"^(?:https?:\/\/)?zenodo\.org\/records\/(\d+)\/?$")
+
+
+def _resolve_doi(doi):
+    r = requests.get(f"https://doi.org/{doi}")
+    r.raise_for_status()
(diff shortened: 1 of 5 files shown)


sandorkertesz:
@chpolste, thanks for adding this new source! 
As for your question:
`Should the test suite have a test that verifies against the actual Zenodo API? If so, I could create a dedicated Zenodo test data record with a few example files.`

Yes, it is important to have a real test like that. It should be protected by

```
@pytest.mark.long_test
@pytest.mark.download
```

Tests marked like that are skipped in the CI and are only active when pytest is run with the `-E long` option. See: https://earthkit-data.readthedocs.io/en/latest/development/tests.html

src/earthkit/data/sources/zenodo.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+class Zenodo(Source):
+    """Source for downloading files from Zenodo records.
+
+    Parameters
+    ----------
+    identifier : int | str
+        Record ID, Zenodo URL or DOI. A DOI may also be given as a doi.org URL.
+    only : str | Sequence[str] | None, optional
```
sandorkertesz, line 70: Maybe `filter` is a better name than`only` for this kwargs. The "file" source already has a `filter` kwarg with a similar meaning.

Issue #936, opened 2026-03-20
Xarray based computations fail after latest code update in xarray engine


sandorkertesz:
### What happened?

When using this code:

```python
#!/usr/bin/env python3

from earthkit import data as ekd

seas5_data = ekd.from_source("url", "https://sites.ecmwf.int/repository/earthkit-data/test-data/seas5_2m_temperature_201501-201503_europe_1deg.grib")
seas5_xr = seas5_data.to_xarray(
    time_dim_mode="forecast", add_valid_time_coord=True
).rename({"2t": "t2m"})   # .compute() #  .compute required for latest ek-d version


seas5_xr_gb = seas5_xr.groupby("valid_time.day")

print(seas5_xr_gb)
```

The following error is generated:

```
 
--- Logging error ---
Traceback (most recent call last):
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test/lib/python3.12/site-packages/earthkit/data/xr_engine/builder.py", line 261, in _raw_indexing_method
    result = r.to_array(index=field_index, array_namespace=self.array_namespace, dtype=self.dtype)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test/lib/python3.12/site-packages/earthkit/data/indexing/tensor.py", line 89, in wrapped
    return func(self, *args, **_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test/lib/python3.12/site-packages/earthkit/data/indexing/tensor.py", line 452, in to_array
    return self._to_array(source_to_array_func, index=index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Caskroom/miniforge/base/envs/test/lib/python3.12/site-packages/earthkit/data/indexing/tensor.py", line 439, in _to_array
    arr, current_field_shape = self._prepare_tensor_data(source_to_array_func, index=index)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #941, opened 2026-03-23
Fix/tensor indexing bug #936

pawel-wolff:
### Description

This PR fixes the issue #936.

The bug was caused by the recent upgrade (https://github.com/ecmwf/earthkit-data/pull/932) of the earthkit tensor indexing support from `BASIC` to `OUTER` (see https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html#indexing-examples).

With the `OUTER` indexing support, xarray can ask the earthkit backend to sub-select a tensor axis using:
* a single integer
* a slice
* **an array of integers or a boolean mask array**

The last kind of indexer turned out to be problematic. More specifically, if a field part `v` of a tensor is a 2D lat-lon array and one wants to subselect it using a pair of indexers of the third kind (e.g. `index=([9, 10, 11], [0, 1])`, see: https://github.com/ecmwf/earthkit-data/blob/8442708e4c3aa95dd7a570fc696188caa8405b04/tests/xr_engine/test_xr_engine_core.py#L226), then the numpy-style indexing `v[index]` (https://github.com/ecmwf/earthkit-data/blob/5d2b2afbdf6da33dd31caa5b8d7acf0eba9f3868/src/earthkit/data/core/field.py#L488)
does not work, because it applies the so-called "advanced indexing" (https://docs.xarray.dev/en/stable/internals/how-to-add-new-backend.html#indexing-examples).

In order to solve this problem, an **outer** indexing is applied explicitly to sub-select a field in the following methods:
* `Field.to_numpy()`: https://github.com/ecmwf/earthkit-data/blob/8442708e4c3aa95dd7a570fc696188caa8405b04/src/earthkit/data/core/field.py#L488
* `Field.to_arrray()`: https://github.com/ecmwf/earthkit-data/blob/8442708e4c3aa95dd7a570fc696188caa8405b04/src/earthkit/data/core/field.py#L534
* `Field.data()`: https://github.com/ecmwf/earthkit-data/blob/8442708e4c3aa95dd7a570fc696188caa8405b04/src/earthkit/data/core/field.py#L618

Note that in the case of similar methods:
* NumpyNDArrayWrapper.to_numpy()
* XarrayDataArrayData.to_numpy()

the indexing for a field is still `v[index]`, so the usual numpy-style indexing applies (which might be an advanced indexing, depending on `index`).

Several tests has been added or extended.

### Contributor Declaration


sandorkertesz:
@pawel-wolff, thank you for this great improvement.

PR #1008, opened 2026-06-02
Feature/parameter component enhancement

pawel-wolff:
### Field

This PR extends `parameter` component of a `Field` by adding the following keys:
* `chem: Optional[str]` (replaces `chem_variable`)
* `chem_long_name : Optional[str]`
* `wavelength: Optional[Union[int, float]]`
* `wavelength_bounds: Optional[Union[tuple[int, int], tuple[float, float]]]`
* `wavelength_units: Optional["Units"]`
* `wave_direction: Optional[float]`
* `wave_direction_index: Optional[int]`
* `wave_direction_bounds: Optional[tuple[float, float]]`
* `wave_direction_units: Optional["Units"]`
* `wave_frequency: Optional[float]`
* `wave_frequency_index: Optional[int]`
* `wave_frequency_bounds: Optional[tuple[float, float]]`
* `wave_frequency_units: Optional["Units"]`

The accessors to the keys representing physical quantities (`wavelength`, `wavelength_bounds`, `wave_direction`, `wave_frequency`, etc.) have now an optional argument `units`; if provided, a unit conversion is performed:
```
f.parameter.wavelength() == 550 # in native units f.parameter.wavelength_units() == "nm"
f.parameter.wavelength("um") == 0.55
```
Note that the following is not yet possible
```
f.get("parameter.wavelength[um]")
```
but can be considered in the future.

The unit conversion is made via `earthkit-utils`, where the appropriate feature was added in the version 1.0.0rc2. For this reason the required version of `earthkit-utils` was changed.

### Xarray engine

Four new built-in dimensions, related to the above `Field` metadata keys, were added to the Xarray engine:
* chem_variable
* wavelength
* wave_direction
* wave_frequency

CF-compliant attributes of the corresponding coordinate variables were proposed (`src/earthkit/data/xr_engine/profiles/defaults.yaml`)

Coordinate sorting was improved: a fallback on sorting string representations is now the very last option, behind sorting homogenous types; see https://github.com/ecmwf/earthkit-data/blob/df5bcdbc0244c715749a068abb2198fc0cc337fa/src/earthkit/data/xr_engine/fieldlist.py#L175
The previous implementation was wrong on sorting float coordinates, because `float` were converted to `str` early (e.g. `"5.0"` > `"11.0"`).

### Mapping parameter keys from GRIB

The metadata key `chem` is mapped from the GRIB key `parameter.chemShortName` (previously this was `chemShortName` which can have `"unknown"` value instead of `None` when undefined).

For mapping the other keys from GRIB see: src/earthkit/data/field/grib/parameter.py

### Mapping parameter keys to GRIB

Currently a limited set of parameter keys is mapped back to GRIB after a user modification (via `Field.set()`, for example):
* `variable` -> `shortName`
* `chem` -> `chemShortName`
* If `wavelength_bounds` is set, then their values (in meters) are mapped to `(firstWavelength, secondWavelength)` and `typeOfWavelengthInterval` is set to 2 (which means "Between first and second limit."). If only `wavelength` is set, then its value (in meters) is mapped to `firstWavelength`. (The complexity of this mapping comes from the fact that a more direct GRIB key `mars.wavelength` is read-only).
* `wave_direction_index + 1` -> `directionNumber`
* `wave_frequency_index + 1` -> `frequencyNumber`

Other keys are not mapped to GRIB for the moment.

### Mapping parameter keys from xarray and mars

Will be addressed in another PR.

### Docs, tests

Documentation, example notebooks and tests were updated, accordingly.

## Contributor Declaration


sandorkertesz:
@pawel-wolff, thanks for this improvement. Please consider to following issues/questions:

- in Parameter `wavelength`, `wave_direction` and `wave_frequency` have no units
- there is no code to write the new keys back to GRIB if they are modified in the field with `set()`. This applies to the Xarray to GRIB conversion too.
- is there an extra cost of getting these extra keys every time a Parameter component is built?

pawel-wolff:
@sandorkertesz , thanks for the comments.

> * in Parameter `wavelength`, `wave_direction` and `wave_frequency` have no units

Now each of these keys have a corresponding key returning a units object, for example: `wavelength_units`. Moreover, it is now possible to access `f.parameter.wavelength(units="um")` to get a wavelength in, say, "um". If desired, this pattern can be reused in Vertical component as well (for `level`).


> * there is no code to write the new keys back to GRIB if they are modified in the field with `set()`. This applies to the Xarray to GRIB conversion too.

A reasonable support for mapping the new keys back to GRIB is now implemented.


> * is there an extra cost of getting these extra keys every time a Parameter component is built?

With
```
fl = ekd.from_source("file", "a-lots-of-small-GRIB-messages.grib").to_fieldlist()
```
the following code
```
for f in fl:
    f.parameter  # this triggers building a Parameter component for a field f
```
run on GRIB data containing regular fields (no chemical/optical/waves) is less than 10% slower comparing to develop branch (i.e. no new parameter keys). This seems to be the cost of querying for a few more GRIB keys (it's already a bit optimised - e.g. if `directionNumber` is not found in a GRIB message, no other related keys are accessed).

src/earthkit/data/field/component/component.py
```diff
@@ -223,6 +222,13 @@ def __setstate__(self, state):
 class SimpleFieldComponent(FieldComponent):
     _KEYS = tuple()
     _ALIASES = dict()
+    # keys accepted by a class __init__(); set by __init_subclass__() for every subclass
+    _ALLOWED_CREATE_KEYS = None
+
+    def __init_subclass__(cls):
+        super().__init_subclass__()
+        sig = inspect.signature(cls.__init__)
+        cls._ALLOWED_CREATE_KEYS = tuple(key for key in sig.parameters.keys() if key not in {"self", "args", "kwargs"})
```
sandorkertesz: This will not work e.g. for ForecastTime  where we can use more keys than the constructor has.

src/earthkit/data/field/component/parameter.py
```diff
@@ -326,19 +678,392 @@ def set(self, *args, **kwargs):
...
+    def __init__(
+        self,
+        variable: str = None,
+        standard_name: str = None,
+        long_name: str = None,
+        units: Union[str, "Units"] = None,
+        chem: str = None,
+        chem_long_name: str = None,
+    ) -> None:
+        Parameter.__init__(self, variable=variable, standard_name=standard_name, long_name=long_name, units=units)
```
sandorkertesz: Should not it be super().__init__(?

src/earthkit/data/field/component/component.py
pawel-wolff: I see the problem...

This was just an attempt to reduce (in the future) the code duplication within the `field/component` subpackage (by providing a generic validator for keys passed to a concrete component class `__init__`, instead of passing an explicit list of keys to `SimpleFieldComponent._normalise_create_kwargs()` and `SimpleFieldComponent._normalise_set_kwargs()`). 

But now I see that it would not be useful for `TimeBase` subclasses, at least. So I will revert the changes and go back to the original design.

src/earthkit/data/field/component/parameter.py
pawel-wolff: Yes, I will change to `super().`

sandorkertesz:
@pawel-wolff, this is a massive improvement! Many thanks!

PR #1021, opened 2026-06-15
Feature/earthkit accessor improved

pawel-wolff:
### Description

This PR solves the problem of a NetCDF serialisation of a spectial "_earthkit" attribute attached to xarray variables.
Since the value of the attribute is a nested dictionary, the solution consists of encoding the dictionary as a JSON-string. A special treatment is made for values of the `bytes` type (`base64` encoding is used).

This solution makes the earthkit-specific metadata persistent in the NetCDF format.

Concerning the API / behaviour changes:
* `[ds|da].earthkit.to_netcdf()` accepts the optional kwarg `remove_earthkit_attrs`, which is `False` by default.

Relevant parts of the documentation and how-to notebooks where updated.

### Contributor Declaration


docs/source/concepts/xarray/overview.rst
```diff
@@ -137,12 +137,26 @@ For further details see the following notebook:
...
+    ds_fl = ekd.from_source("sample", "pl.grib")
+    ds_xr = ds_fl.to_xarray()
+    ds_xr.to_netcdf("_from_grib.nc")
+
 
 Using the earthkit accessor
 ++++++++++++++++++++++++++++
 
-By default, the earthkit Xarray engine attaches some special attributes to the generated Xarray dataset that cannot be written to NetCDF. In order to make ``to_netcdf()`` work we need to invoke it on the ``earthkit`` accessor and not directly on the Xarray dataset.
+It is also possible to call ``to_netcdf()`` on the ``earthkit`` accessor. This is equivalent to calling it directly on the dataset.
```
sandorkertesz: Actually now there is a slight difference. This will not write the `_earthkit` attribute to NetCDF (the encoder removes it). However, 
`xarray.Dataset.to_netcdf()` with the new code will !

src/earthkit/data/xr_engine/engine.py
```diff
@@ -12,6 +12,8 @@
 import xarray
 from xarray.backends import BackendEntrypoint
 
+from earthkit.data.xr_engine.accessor import ACCESSOR_KEY
```
sandorkertesz: I would only use local import in `engine.py`. If there is an error in the imported earthkit code Xarray becomes unusable because it tries to import the engine on startup even if the the user does not intend to use earthkit at all.

docs/source/concepts/xarray/overview.rst
pawel-wolff: The default behaviour of `ds.earthkit.to_netcdf()` was already changed (to get the previous behaviour one has to pass kwargs `remove_earthkit_attrs=True`).

The old behaviour was still in force in `.to_target()`: when using `.to_xarray(**kwargs)`, the option `kwargs["add_earthkit_attrs"] = False` was injected. Now both for NetCDF and Zarr `kwargs` are passed unaltered into the xarray engine.

src/earthkit/data/xr_engine/engine.py
pawel-wolff: OK, got it!

PR #1038, opened 2026-06-24
handle optional dependencies in tests

colonesej:
This pull request improves test robustness and reliability by adding conditional skips to tests that depend on optional modules such as `geopandas` and `pyodc`. It introduces new flags in the testing utilities to check for the presence of these dependencies and updates relevant tests to skip execution gracefully if the required modules are missing, preventing unnecessary test failures.

**Test robustness and conditional skipping:**

* Added `NO_ODC` and `NO_GEOPANDAS` flags in `earthkit.data.utils.testing` to check for the presence of `pyodc` and `geopandas` modules, respectively.
* Updated ODB-related tests to skip at the module level if `pyodc` is not installed, using `NO_ODC` and `pytest.skip`. [[1]](diffhunk://#diff-9a8bde3b4250a7a5792621a31396eb9b3b722c056a308feeb5c8f70214831f34R12-R18) [[2]](diffhunk://#diff-f53f31bb4ba34c04cdc4a108eef40c0bfca5caff5cc7387e9b76df1f8a2612a7R13-R19) [[3]](diffhunk://#diff-0778301168a0b46f00010e4aa108ba5a16aaa66e69a6270ebe4249a21f30f051R12-R18) [[4]](diffhunk://#diff-606927025c3e221934324fe68560e9c9734fa304089fb6eec27a56250e84681fL20-R20) [[5]](diffhunk://#diff-606927025c3e221934324fe68560e9c9734fa304089fb6eec27a56250e84681fR136)
* Updated GeoPandas-related tests to skip if `geopandas` is not installed, using `NO_GEOPANDAS` and `pytest.mark.skipif` or `pytest.skip`. [[1]](diffhunk://#diff-225a707dd602b67d225cba1f0181cf3d5e79daad829ced5fe747ebed8d380e37R12-R18) [[2]](diffhunk://#diff-92d5494ce781adf8e2ef28822787c3a6490d4db19c26d9ac5217de410e9ee32aR12-R18) [[3]](diffhunk://#diff-fc0976c2f008849059dff68ceacc9a5172e2c6881db74a71cc2ae29deccfa0c1L14-R17) [[4]](diffhunk://#diff-ffa93bfa4bce30b936bb7cff4207e89167815dc13265f26826ad8aea83e599a5R12-R19) [[5]](diffhunk://#diff-872e49bea3dc0f5c20eddce100e721037a0edbc6d46df0dca0ae738ca98b842dL15-R19) [[6]](diffhunk://#diff-872e49bea3dc0f5c20eddce100e721037a0edbc6d46df0dca0ae738ca98b842dR46-R49) [[7]](diffhunk://#diff-6644451f5507aab563c71a07173b6fc5631282873af6ae425a36fc31a319fdb5R12-R15)

These changes ensure that tests requiring optional dependencies do not fail unnecessarily, improving the reliability of the test suite across different environments.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/utils/testing.py
+++ b/src/earthkit/data/utils/testing.py
@@ -134,6 +134,7 @@ if not NO_PROD_FDB:
     fdb_home = os.environ.get("FDB_HOME", None)
     NO_PROD_FDB = fdb_home is None
 
+NO_ODC = not modules_installed("pyodc")
 NO_GRIBJUMP = NO_FDB or not modules_installed("pygribjump")
 NO_POLYTOPE = not os.path.exists(os.path.expanduser("~/.polytopeapirc"))
 NO_COVJSONKIT = not modules_installed("covjsonkit")
@@ -141,6 +142,8 @@ NO_RIOXARRAY = not modules_installed("rioxarray")
 
 NO_S3_AUTH = not modules_installed("aws_requests_auth")
 NO_GEO = not modules_installed("earthkit-geo")
+NO_GEOPANDAS = not modules_installed("geopandas")
+
 try:
     NO_ECFS = not os.path.exists(shutil.which("ecp"))
 except Exception:
--- a/src/earthkit/data/utils/testing.py
+++ b/src/earthkit/data/utils/testing.py
@@ -134,7 +134,7 @@ if not NO_PROD_FDB:
     fdb_home = os.environ.get("FDB_HOME", None)
     NO_PROD_FDB = fdb_home is None
 
-NO_ODC = not modules_installed("pyodc")
+NO_ODC = not modules_installed("pyodc") and not modules_installed("codc")
 NO_GRIBJUMP = NO_FDB or not modules_installed("pygribjump")
 NO_POLYTOPE = not os.path.exists(os.path.expanduser("~/.polytopeapirc"))
 NO_COVJSONKIT = not modules_installed("covjsonkit")
--- a/tests/high_level_object/test_hl_geojson_core.py
+++ b/tests/high_level_object/test_hl_geojson_core.py
@@ -9,10 +9,13 @@
 # nor does it submit to any jurisdiction.
 #
 
+import pytest
+
 from earthkit.data import from_source
-from earthkit.data.utils.testing import earthkit_test_data_file
+from earthkit.data.utils.testing import NO_GEOPANDAS, earthkit_test_data_file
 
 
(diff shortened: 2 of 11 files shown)


src/earthkit/data/utils/testing.py
```diff
@@ -134,13 +134,16 @@ def modules_installed(*modules):
     fdb_home = os.environ.get("FDB_HOME", None)
     NO_PROD_FDB = fdb_home is None
 
+NO_ODC = not modules_installed("pyodc")
```
Oisin-M: I think may need to check if `codc` also available

colonesej: ~isn't it a `pyodc` dependency?~

from the pyodc docs 
```
pyodc is a pure-python encoder and decoder for ODB-2 data, which encodes data from, and decodes it into pandas data frames
codc is an implementation of the same API as pyodc that depends on the ECMWF odc library, and comes with much better performance.
```
So I would guess that pyodc is the fallback in case you do not have codc !?. I based this for the tests that only import `pyodc`

in the ODBReader we kind of have this behaviour

```python
class ODBReader(Reader):
    _format = "odb"

    def to_pandas(self, odc_read_odb_kwargs=None):
        try:
            import codc as odc
        except Exception:
            try:
                import pyodc as odc
            except ImportError:
                raise ImportError("ODC handling requires 'pyodc' to be installed")
            LOG.debug("Using pure Python odc decoder.")

```
I can add it but it may not be necessary right now!? What do you think?

Oisin-M: Edit: sorry, was typing up and then saw you already responded!

> So I would guess that pyodc is the fallback in case you do not have codc !?

Yes, it's much less performant so `codc` is preferred

Oisin-M: What about just changing to
```suggestion
NO_ODC = not modules_installed("pyodc") and not modules_installed("codc")
```
?

colonesej: I still think it's too restrictive because this variable is only used for testing. And let's say you don't have a wheel for codc on your system. It shouldn't prevent from running tests against `pyodc`

Oisin-M: > I still think it's too restrictive because this variable is only used for testing. And let's say you don't have a wheel for codc on your system. It shouldn't prevent from running tests against pyodc

it doesn't - if you have either or both of pyodc or codc, the tests would run. They only wouldn't run if you didn't have either library

Oisin-M (approved):

PR #1039, opened 2026-06-24
Fix/level types

pawel-wolff:
### Description

Definitions of level types and their mapping with GRIB changed:
* whenever GRIB level type does not carry information in `level`, `bottomLevel`, `topLevel` (which implies they are all 0), the class `GribSurfLevelType` is used;
* a level type is a layer (and the corresponding GRIB level type is `GribLayerType`) only when both the bottom and top surfaces are defined (via `bottomLevel` / `topLevel`); for examples, `"medium_cloud_layer"` is a layer (defined by a tuple bottom/top in hPa), but not `"low_cloud_layer"` nor `"high_cloud_layer"` (which are defined by a single - upper or lower interface, in hPa).

Dimensionless units are defined as `None`, which is correctly recognised by pint and taken care when building CF units.
(The CF dimensionless units "1" are not well handled by pint).

Bottom and top now respect the geometry. A `Field.vertical.layer()` is `(bottom, top)`, if defined.

Whenever the attribute `"standard_name"` of a level type was not an existing CF standard name (cf. https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html), it has been removed.

Tests adapted.

### Contributor Declaration


src/earthkit/data/field/grib/vertical.py
```diff
@@ -156,8 +171,9 @@ def to_grib(self, component, handle=None):
         # this tries to avoid writing the coefficients back to the handle if they are already
         # present and correct, which can be expensive for large coefficient arrays. The check
         # is not robust enough and should be improved.
+        # TODO: should be: if handle is not None and hasattr(component, "_level_parameters"): ?
         if handle is not None and hasattr(component, "level_parameters"):
-            level_parameters = component._level_parameters
```
sandorkertesz, line 160: Yes, this is old code. I will correct it.

PR #1072, opened 2026-07-16
Fix retain time coordinate

Alfie5640:
### Description

Updated default profiled to enable add_valid_time_coord
When add_valid_time_coord=True, converting GRIB data containing a single valid time could fail during xarray conversion.

The failure occurred because scalar "valid_time" values were incorrectly assigned a coordinate dimension: ValueError: dimensions ('valid_time',) must have the same length as the number of data dimensions, ndim=0

This change updates coordinate dimension inference so scalar coordinate values are represented with no dimensions ( dims=() ), while non-scalar coordinates retain their inferred dimensions

### Tests

Added a regression test covering the single valid time case to make sure that

- valid_time is retained as an auxiliary coordinate
- valid_time is not incorrectly promoted to a dimension
- scalar coordinate dimensions are represented correctly

Ran all tests with pytest

### Contributor Declaration


src/earthkit/data/xr_engine/profiles/defaults.yaml
```diff
@@ -23,7 +23,7 @@ variable_attrs: []
 global_attrs: []
 add_earthkit_attrs: true
 rename_attrs: {}
-add_valid_time_coord: false
+add_valid_time_coord: true
```
pawel-wolff: I don't think we should change the default options, as it would change the existing behavior and could be a breaking change.

tests/xr_engine/test_xr_engine_incomplete_tensor.py
pawel-wolff, line 1: With the option `add_valid_time_coord` back to the default `False` the tests changed here require the option `add_valid_time_coord=True` (or parametrisation it with `[True, False]`) in order to effectively test the fix.

pawel-wolff (changes requested):
@Alfie5640, thanks for finding and fixing this bug! Please take a look at my comments.

Alfie5640:
@pawel-wolff Thanks for the review
I reverted the default add_valid_time_coord option to keep existing behaviour and updated the tests to enable add_valid_time_coord=True. The full test suite still passes locally

pawel-wolff:
Thank you! Please sync with the `develop` branch to go ahead with it.

PR #1076, opened 2026-07-16
Add regression test for MultiData.to_pandas via CDS multi-source

Alfie5640:
## Description

MultiData.to_pandas was unimplemented (just pass), so calling .to_pandas() on data returned by multi-source CDS requests (e.g. `sis-european-wind-storm-reanalysis`) returned None instead of a Dataframe

While testing this against a real CDS request, found a second related bug: mergers/pandas.py::merge was calling .to_pandas() on raw Source objects, which don't implement that method, only the result of `.to_data_object()` does. Fixed by calling `.to_data_object().to_pandas()` for each source

- src/earthkit/data/data/multi.py: MultiData.to_pandas now delegates to self.sources.to_pandas(...)
- src/earthkit/data/mergers/pandas.py: merge each source via to_data_object().to_pandas() instead of to_pandas() directly

Added test_cds_multi_source_to_pandas:
- Fails against un-fixed branch with AttributeError: 'NoneType' object has no attribute empty
- Passes with this fix applied

Ran tests with pytest (no failures)

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/data/multi.py
+++ b/src/earthkit/data/data/multi.py
@@ -105,7 +105,8 @@ class MultiData(SimpleData):
         :py:class:`pandas.DataFrame`
             A Pandas DataFrame containing data from all sources.
         """
-        pass
+        
+        return self.sources.to_pandas(*args, **kwargs)
 
     def to_geopandas(self, *args, **kwargs):
         """Convert into a GeoPandas GeoDataFrame.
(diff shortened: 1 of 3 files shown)


sandorkertesz:
@Alfie5640, please see #1075, which already addresses this issue.

Alfie5640:
Thanks for pointing that out. I'll close this PR since #1075 addresses the underlying issue

PR #1086, opened 2026-07-27
Normalized ecCodes from unknown to None

Alfie5640:
### Description
Problem

When converting GRIB fields to xarray via to_xarray(), several common ERA5 variables (2t, 10u, 10v, tp) end up with standard_name="unknown" in their attrs. This happens because ecCodes returns the literal string "unknown" for cfName when no CF mapping is resolved, and this value was passed through unfiltered into the field's standard_name. Since "unknown" is not a valid CF standard name, this misleadingly implies a resolved CF mapping exists when it doesn't.

Fix

In GribParameterBuilder._build_dict (src/earthkit/data/field/grib/parameter.py), cfName == "unknown" is now normalized to None. Downstream, the xr_engine attrs pipeline already drops None values before writing final attrs, so standard_name is now correctly omitted for these fields rather than containing an invalid literal.

Scope

This is a minimal, source-level fix. The issue also raised the possibility of falling back to a cfVarName-derived CF standard_name mapping, but I've scoped that out of this PR since it's a separate design decision. Happy to open a follow-up issue for that if maintainers want to pursue it.

Testing

Reproduced the exact scenario from the issue (CDS ERA5 fetch of 2t, 10u, 10v, msl, sp, tp) and confirmed all 4 affected variables now correctly omit standard_name, while msl/sp (already correctly resolved) are unaffected.
Updated tests/grib/test_grib_parameter.py (2 assertions) and tests/xr_engine/test_xr_engine_dims.py (3 expected-attrs dicts) that had encoded the old buggy "unknown" value as expected behavior.

Full test suite passes

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/field/grib/parameter.py
+++ b/src/earthkit/data/field/grib/parameter.py
@@ -46,6 +46,11 @@ class GribParameterBuilder:
             v = _get("param", None)
         variable = v
         standard_name = _get("cfName", None)
+
+        if standard_name == "unknown":
+            # ecCodes may return "unknown" for unresolved cfName. None is safer downstream
+            standard_name = None
+
         long_name = _get("name", None)
 
         units = _get("units", None)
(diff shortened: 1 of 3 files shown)


pawel-wolff (approved):
@Alfie5640, thank you for fixing this!

Issue #1037, opened 2026-06-24
Some API inconsistencies to consider


Oisin-M:
Thought would be good to flag before 1.0 in case we want to do anything.

- `from_source("fdb", ../)` has a `userconfig` kwarg whereas would expect `user_config`
- `from_source("wekeocds", ...)` whereas would expect `from_source("wekeo-cds", ...)`

Alfie5640:
Since earthkit-data is now at 1.0, would an acceptable approach be adding aliases (user_config/wekeo-cds) while keeping backwards compatibility, rather than renaming the existing API? 
Also, are `userconfig` and `wekeocds` intended as specific examples to address, or is the goal to do a wider API naming consistency review?

Oisin-M:
> Since earthkit-data is now at 1.0, would an acceptable approach be adding aliases (user_config/wekeo-cds) while keeping backwards compatibility, rather than renaming the existing API?

Yes, any breaking changes such as this would now need a major release and adding aliases and then slow deprecation would be way to do this.

> Also, are userconfig and wekeocds intended as specific examples to address, or is the goal to do a wider API naming consistency review?

I opened this when reviewing some of the API. I think those are two examples which are a bit inconsistent with most of the other naming, but there are surely others. Happy for the issue to grow - eventually a wider API naming consistency review would be nice

PR #1091, opened 2026-08-01
Added aliases for user_config (fdb) and wekeo-cds (wekeocds) with deprecation warnings

Alfie5640:
Addresses the two examples raised in #1037:

- `from_source("fdb", ...)` now accepts `user_config` as well as the existing `userconfig` kwarg. `userconfig` still works but raises a FutureWarning pointing to `user_config`.
- `from_source("wekeo-cds", ...)` now works alongside the existing `from_source("wekeocds", ...)`. The old "wekeocds" name still works but raises a FutureWarning pointing to "wekeo-cds".

As pointed toward on the discussion on the issue, this is backward compatible only, so  no breaking changes. Both old names continue to work until a future major version removes them.

## Implementation notes

- The `fdb` alias is a straightforward kwarg addition in `FDBSource.__init__`, with the deprecation warning fired when the old name is passed. The underlying `pyfdb.FDB(**kwargs)` call still receives `userconfig` internally, since that's the FDB library's own kwarg name. (only earthkit's user facing parameter is renamed)
- The `wekeocds` alias required a small addition to `SourceMaker` (`sources/__init__.py`) rather than just a kwarg: an `ALIASES` dict maps `"wekeo-cds"` to `"wekeocds"` so the existing `wekeocds.py` source file can be resolved under either name without renaming the file itself. `wekeocds.py`'s registered `name` attribute is left untouched.
- Also added the corresponding `@overload` for `Literal["wekeo-cds"]` and `user_config` in the `from_source` type-hint block, so both names get proper autocomplete/type checking.

## Audit

Also did a pass over every source in `src/earthkit/data/sources/` (constructor kwargs and registered source names) looking for other cases of the same two inconsistency patterns (hyphenation in source names, missing underscores in kwargs). Didn't find any others in these files, so these two seem to be outliers rather than telling of a wider problem here.

## Tests

- New `tests/sources/test_source_aliases.py`. Dependency free tests covering the `SourceMaker` alias/warning logic directly (no `pyfdb`/`hda` required, so these run everywhere)
- New tests in `tests/sources/test_fdb.py` (`test_fdb_userconfig_deprecated_alias`, `test_fdb_user_config_no_warning`) covering the `userconfig`/`user_config` deprecation behavior in `FDBSource`, gated with the same `NO_FDB` skip marker as the files existing tests

Leaving #1037 open,  to keep it open for a possible wider API naming review

### Contributor Declaration


src/earthkit/data/sources/fdb.py
```diff
@@ -26,19 +27,29 @@
...
         super().__init__()
 
         for k in ["group_by", "batch_size"]:
             if k in kwargs:
                 raise ValueError(f"Invalid argument '{k}' for FDBSource. Deprecated since 0.8.0.")
 
+        if userconfig is not None:
+            warnings.warn(
+                "'userconfig' is deprecated, use 'user_config' instead",
+                FutureWarning,
```
sandorkertesz: I wonder if it should be a `DeprecationWarning` instead.

src/earthkit/data/sources/fdb.py
```diff
@@ -26,19 +27,29 @@
...
+    def __init__(
+        self, *args, request=None, stream=True, config=None, userconfig=None, user_config=None, lazy=False, **kwargs
+    ):
         super().__init__()
 
         for k in ["group_by", "batch_size"]:
             if k in kwargs:
                 raise ValueError(f"Invalid argument '{k}' for FDBSource. Deprecated since 0.8.0.")
 
+        if userconfig is not None:
```
sandorkertesz, line 42: I think if both `userconfig` and `user_config` are specified an exception should be raised.

sandorkertesz:
@Alfie5640, thank you for adding this improvement. Please see my comments above.

Alfie5640:
Thanks for the review! I've updated the FDB handling to use DeprecationWarning, added an error when both userconfig and user_config are provided, and updated the tests accordingly. I left the wekeocds warning unchanged for now since it wasn't part of the review but I can change if needed

PR #1098, opened 2026-08-14
Fix: Temporary patch for ORCA grid shape swapping

sandorkertesz:
### Description

Problem

With `eckit 2.1.0.21`, the shape of all ORCA grids is returned with swapped dimensions. This causes ORCA grids to be unusable with `earthkit-data` until the issue is resolved upstream in `eckit`.

Solution

A temporary patch has been added to `earthkit-data` to work around the eckit bug:

- Automatic detection: `earthkit-data` now automatically detects whether an ORCA grid shape is in the incorrect order.
- Conditional swap: The dimensions are swapped only when the incorrect ordering is detected.
- Future-proof: The fix is scoped to the currently supported ORCA grids, ensuring that future changes to ORCA grid handling in eckit will not be inadvertently affected.

Notes

This is a temporary workaround and should be removed once the upstream fix is released in eckit.


src/earthkit/data/field/component/geography.py
```diff
@@ -909,6 +911,13 @@ def shape(self) -> tuple:
         -------
         tuple
         """
+        grid_type = self._grid.type
+        # temporary fix for ORCA grids, which are returned Grid as (Ni, Nj) instead of (Nj, Ni)
+        if grid_type and "ORCA" in grid_type:
+            shape = self._grid.shape
+            if shape is not None and shape in _SWAPPED_ORCA_GRID_SHAPES:
+                shape = (shape[1], shape[0])
+                return shape
```
pmaciel, line 920: There is no other ORCA grids except those in the metadata yaml, so it is safe to (as a temporary fix) directly reverse the shape order (say, using reversed()), without requiring a pre-known list of shapes

sandorkertesz: This code will even work when this is fixed in eckit (or new orca grids are added in the future). So using a pre-defined list is the safest solution.

Issue #1097, opened 2026-08-14
Raise when invalid `kwargs` passed to `from_source`


Oisin-M:
### What happened?

For many methods, passing invalid kwargs just leads to the kwargs being swallowed with no warning to the user. This makes it quite difficult to be sure code is working as intended - and exacerbated by any vibe-coding since AIs are extremely good at making up random kwargs that sound plausible.

### What are the steps to reproduce the bug?

Here is a basic example:
```python
# actual kwarg is lazily=True, but no warning or anything emitted here so hard for user to know
i_probably_think_this_is_a_lazy_source = ekd.from_source("file", "data.grib", lazy=True)
```

### Version

current

### Platform (OS and architecture)

//

### Relevant log output

```shell

```

Oisin-M:
A solution I would like is to split out the different source kwargs explicitly, with signatures of what we have in the docs: https://earthkit-data.readthedocs.io/en/latest/concepts/inputs/from_source.html.

Could then have something along the lines of:
```python
POSSIBLE_SOURCES = {
   "file": _from_file,
   ...
}

def from_source(source, *args, **kwargs):
   return POSSIBLE_SOURCES[source](*args, **kwargs)

def _from_file(path: str,
    expand_user: Union[bool, list, tuple] = True,
    expand_vars: bool = False,
    unix_glob: bool = True,
    recursive_glob: bool = True,
    filter: Union[str, Callable] = None,
    parts: list = None,
    stream: bool = False,
) -> "Data": ...
```

This wouldn't change any of the external API, but I think would make it easier to track all these arg/kwarg issues and may make documentation easier also

Oisin-M:
This is resolved for some sources, namely "file" via #1112, but not for all (request-type sources not resolved yet)

PR #1112, opened 2026-08-19
Feat/raise invalid kwarg from source

Oisin-M:
### Description
Work towards #1097...

My main aim is to have sources and readers error early upon instantiation when invalid kwargs passed. For sources which mutate, should then pass kwargs along down the chain, but at some point at the end of the mutations the list of valid kwargs should be concrete and should directly error.

Edit: decided not erroring but warning instead.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/readers/covjson/__init__.py
+++ b/src/earthkit/data/readers/covjson/__init__.py
@@ -29,7 +29,7 @@ def reader(source, path, *, magic=None, deeper_check=False, content_type=None, *
     def _reader():
         from .reader import CovJSONReader
 
-        return CovJSONReader(source, path)
+        return CovJSONReader(source, path, **kwargs)
 
     if _match_content_type(content_type) or _match_magic(magic, deeper_check):
         return _reader()
--- a/src/earthkit/data/readers/covjson/__init__.py
+++ b/src/earthkit/data/readers/covjson/__init__.py
@@ -35,7 +35,7 @@ def reader(source, path, *, magic=None, deeper_check=False, content_type=None, *
         return _reader()
 
     extension = pathlib.Path(path).suffix
-    if extension in [".covjson"]:
+    if extension == ".covjson":
         return _reader()
 
     kind, _ = mimetypes.guess_type(path)
--- a/src/earthkit/data/readers/covjson/__init__.py
+++ b/src/earthkit/data/readers/covjson/__init__.py
@@ -61,10 +61,6 @@ def stream_reader(
     **kwargs,
 ):
     if _match_content_type(content_type) or _match_magic(magic, deeper_check):
-        # if memory:
-        #     from .reader import CovjsonMemoryReader
-
-        #     return CovjsonMemoryReader._from_stream(stream)
         from .reader import CovJSONStreamReader
 
         return CovJSONStreamReader(stream)
--- a/src/earthkit/data/readers/covjson/__init__.py
+++ b/src/earthkit/data/readers/covjson/__init__.py
@@ -29,11 +29,7 @@ def reader(source, path, *, magic=None, deeper_check=False, content_type=None, *
     def _reader():
         from .reader import CovJSONReader
 
-<<<<<<< HEAD
         return CovJSONReader(source, path, **kwargs)
-=======
-        return CovJSONReader(source, path)
->>>>>>> e9d871db (chore: renaming)
 
     if _match_content_type(content_type) or _match_magic(magic, deeper_check):
         return _reader()
@@ -67,12 +63,6 @@ def stream_reader(
     if _match_content_type(content_type) or _match_magic(magic, deeper_check):
         from .reader import CovJSONStreamReader
 
-<<<<<<< HEAD
-=======
-        #     return CovjsonMemoryReader._from_stream(stream)
-        from .reader import CovJSONStreamReader
-
->>>>>>> e9d871db (chore: renaming)
         return CovJSONStreamReader(stream)
 
 
--- a/src/earthkit/data/readers/covjson/__init__.py
+++ b/src/earthkit/data/readers/covjson/__init__.py
@@ -29,7 +29,11 @@ def reader(source, path, *, magic=None, deeper_check=False, content_type=None, *
     def _reader():
         from .reader import CovJSONReader
 
+<<<<<<< HEAD
         return CovJSONReader(source, path, **kwargs)
+=======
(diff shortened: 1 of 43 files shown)


src/earthkit/data/readers/covjson/__init__.py
```diff
@@ -26,21 +26,18 @@ def _match_magic(magic, deeper_check):
 
 
 def reader(source, path, *, magic=None, deeper_check=False, content_type=None, **kwargs):
-    def _reader():
-        from .reader import CovjsonReader
-
-        return CovjsonReader(source, path)
+    from .reader import CovjsonReader
```
sandorkertesz: We want to avoid this import upfront.

Oisin-M: Will revert, but why do we want to avoid here? It doesn't import covjsonkit or any optional dependencies and doesn't seem to do any large work that I can see

Oisin-M:
@sandorkertesz, I've decided that any refactoring of the `from_source` logic should be a separate PR and also isn't needed to solve the `kwargs` issue so I've marked ready for review.

This is working for file type source. I have not tested in detail every other configuration e.g. dummy sources etc.

Oisin-M:
This issue gets much trickier for some other sources. For example, ecmwf open-data `EODRetriever` passes the `kwargs` to the `RequestBuilder` and it doesn't pass it down the chain. So e.g. `grib_handle_policy` wouldn't have any effect i.e.
```python
ekd.from_source(
        "ecmwf-open-data",
        param=["2t", "msl"],
        levtype="sfc",
        grib_handle_policy=grib_handle_policy
    )
>> The following keywords {'grib_handle_policy'} are ignored
>> GribData(path=...)
```
This is all due to the fact we allow passing requests via kwargs, so there is a fundamental conflict and inconsistency between if a kwarg should be passed to the source handling part of the code, or the reader part of the code, or for handling if streaming/lazy etc. This is already a problem for the kwarg `stream` and the fdb and mars sources. I will open an issue with this separately, but I think we need to think how we want to handle that in a sensible way

Oisin-M:
Noting that raising directly is too naive an approach since we can retrieve folders etc. and may want to still specify grib policy for grib files, even if not all files are grib

src/earthkit/data/readers/covjson/__init__.py
sandorkertesz: You are right, it is safe to import CovjasonReader, so please ignore my comment.

src/earthkit/data/readers/covjson/file.py
```diff
@@ -75,6 +75,5 @@ def describe(self, *args, **kwargs):
         pass
 
     def to_data_object(self):
-        from earthkit.data.data.geojson import GeoJsonData
 
-        return GeoJsonData(self)
+        return CovJSONList(self)
```
sandorkertesz: I think the correct object to return here is CovJSONData. The previous behaviour of returning  GeoJsonData was a bug.

Oisin-M: Good catch

sandorkertesz:
@Oisin-M, I wonder if it would be possible to split this PR into two separate PRs: one with the renaming and one with the actual changes. Then it would be easier to review these potentially critical changes.

Oisin-M:
> @Oisin-M, I wonder if it would be possible to split this PR into two separate PRs: one with the renaming and one with the actual changes. Then it would be easier to review these potentially critical changes.

Yes, I will update and split into separate PRs. Renaming is now in #1112. I will keep the other logic here

Oisin-M:
@sandorkertesz this is now showing the correct diff without all the renaming. This causes exceptions to be raised when any kwargs are passed to readers that they don't recognise. Because of multifile sources and backwards compatibility, we could instead just warn that the kwargs are being ignored for now if that's better? This only applies for file-based sources where the kwargs get passed down, it does not apply for request-based sources due to the way kwargs are routed directly to the requests

src/earthkit/data/sources/multi.py
```diff
@@ -157,7 +157,7 @@ def _call(s, *args, **kwargs):
         from earthkit.data.utils.progbar import tqdm
 
         with SoftThreadPool(nthreads=nthreads) as pool:
```
sandorkertesz, line 159: The callable s can use the observer 'kwarg'. There is nothing in ekd that would use that, so this seems to be a legacy kwarg. However, there can be plugin sources or sources implemented in another packages (e.g. anemoi) with arbitrary kwargs, they might utilise this.

Oisin-M: I suppose it's possible, but I wasn't considering it as part of the public API. If you think it's better, I can explicitly add it as a kwarg to all the sources where it's passed along to and we deprecate slowly, but otherwise I would vote to remove

sandorkertesz: It is worth checking if it is used anywhere in anemoi. Otherwise I agree it could be removed, it is better then adding it to all the sources (as deprecated).

sandorkertesz:
> @sandorkertesz this is now showing the correct diff without all the renaming. This causes exceptions to be raised when any kwargs are passed to readers that they don't recognise. Because of multifile sources and backwards compatibility, we could instead just warn that the kwargs are being ignored for now if that's better? This only applies for file-based sources where the kwargs get passed down, it does not apply for request-based sources due to the way kwargs are routed directly to the requests

I would vote for issuing a warning, that is the safest solution.

Oisin-M:
> > @sandorkertesz this is now showing the correct diff without all the renaming. This causes exceptions to be raised when any kwargs are passed to readers that they don't recognise. Because of multifile sources and backwards compatibility, we could instead just warn that the kwargs are being ignored for now if that's better? This only applies for file-based sources where the kwargs get passed down, it does not apply for request-based sources due to the way kwargs are routed directly to the requests
> 
> I would vote for issuing a warning, that is the safest solution.

I have now updated to raise a UserWarning, and I have added tests. This works correctly for the file source default readers (not streaming, not in memory), and it doesn't quite work for the zipfiles yet because they do not forward kwargs correctly at the moment. This is an existing bug and unrelated to this PR so I have opened a different issue here: #1137

src/earthkit/data/sources/multi.py
Oisin-M: I have looked for the word "observer" in the codebases' default branches and have found no matches

PR #1127, opened 2026-08-27
Feat/simplify readers

Oisin-M:
### Description
Was getting confused by the logic in readers so opening PR to hopefully simplify a few things.

Couple of things I note
- probe_csv checking in zip files etc. can be removed since it can never happen, because zip files will be read by the zip reader => removed
- txt file check could read unbounded amount of bytes => updated
```python
 with open(path, "r", encoding="utf-8") as f:
            for i, _ in enumerate(f):
                if i > prob_lines:
                    break
```
- not addressed but readers read in alphabetical order is **required** at the moment because the `magic is None` case is not handled correctly for many readers which are alphabetically after the `directory` reader which will match to it. Would be nice to maintain a priority order later that isn't alphabetical

### Contributor Declaration


src/earthkit/data/sources/ecmwf_api.py
```diff
@@ -109,10 +109,4 @@ def to_pandas(self, **kwargs):
             pandas_read_csv_kwargs=pandas_read_csv_kwargs,
             # odc_read_odb_kwargs=odc_read_odb_kwargs,
             **kwargs,
-        )
-
-    def empty_reader(self, *args, **kwargs):
-        if self.expect_any:
```
sandorkertesz, line 115: This should not be removed. "expect=any" in the request has a specific meaning. As a consequence, the code utilising "empty_readers" should be kept under readers. Also see these tests: test_mars_grib_expect_any_1, test_mars_grib_expect_any_2

src/earthkit/data/readers/text.py
```diff
@@ -38,15 +39,6 @@ def ignore(self):
         # Used by multi-source
         return True
 
-    def mutate(self):
```
sandorkertesz, line 41: Why was this removed?

Oisin-M: If it's a CSV from what I understand then it shouldn't ever get here and should be in the CSV reader, so the whole `is_csv` check is not needed. If that's the case, then we just always return self and therefore I removed the entire mutate function so it just inherits the default mutate which just returns self

src/earthkit/data/sources/ecmwf_api.py
Oisin-M: Will readd. For my general understanding, is this intended to be extended later to arbitary sources? It currently only applies for the ecmwf-api class, which mars then inherits from. And what's the reason to return an empty reader vs raising an exception?

sandorkertesz: This is MARS specific, see: https://confluence.ecmwf.int/spaces/UDOC/pages/118817963/expect+-+keyword+in+MARS+Dissemination+request .  Particularly, if "expect=any" is used and no data is available MARS does not fail. In line with this ekd returns an empty object instead of failing. This behaviour was requested by a user to be fully in line with the MARS behaviour.

src/earthkit/data/readers/text.py
sandorkertesz: At this level we do not know in what order the readers are tested. So in theory it can happen that the text reader is tested before the csv one.

Oisin-M: From my understanding the readers are always checked in alphabetical order via this line here:
https://github.com/ecmwf/earthkit-data/blob/87a80031a49ba8db22da1e950c18f9128a081206/src/earthkit/data/readers/__init__.py#L131

sandorkertesz: That is true. I think a comment should be added somewhere that whatever the order is the Text reader should only be evaluated after the CSV reader. Probably this should be enforced by adding priority to the evaluation  (future development).

sandorkertesz:
@Oisin-M, FYI, the empty source changed in another  PR: #1132

src/earthkit/data/readers/text.py
Oisin-M: It's semi-enforced by only running the text check when `deeper_check is True`, but to be honest a better way to solve this would just be to maintain a priority and not need the deep/shallow distinction

Oisin-M:
`tests/readers/test_csv_reader.py::test_csv_text_file` fails because the detection of csv is based upon `kind, compression = mimetypes.guess_type(path)`. Could change to be based upon opening the actual data instead if that's preferable, but it's potentially slow. Could also instead have the checking the full data happen in the deeper check as a solution also. What do you prefer as a solution?

Edit: did a workaround checking a few bytes like we do for the txt reader

Issue #1070, opened 2026-07-15
MultiFile Data is difficult to use


HCookie:
### What happened?

When opening a tar archive and enabling the `tarfile` check, it returns a `earthkit.data.data.multi.MultiData` object which to the best of my ability I cannot figure how to use, and the `.sources` property has some issues.
```
>>> ds.sources[0]
Traceback (most recent call last):
  File "<python-input-20>", line 1, in <module>
    ds.sources[0]
    ~~~~~~~~~~^^^
  File "/.../earthkit/data/sources/multi.py", line 75, in __getitem__
    while n >= self._length(i):
               ~~~~~~~~~~~~^^^
  File "/.../earthkit/data/sources/multi.py", line 94, in _length
    self._lengths[i] = len(self.sources[i])
                       ~~~^^^^^^^^^^^^^^^^^
TypeError: object of type 'File' has no len()
```
This tar archive is made from multiple `csv's`, or at least `ekd` thinks it is.
How do I use this object? Or what can we add to improve it?

### Platform (OS and architecture)

ATOS

### Relevant log output

```shell

```

sandorkertesz:
@HCookie, can you share with me your input data? 

The MultiData object is used to represent a collection of files that earthkit cannot merge into a single object. The `available_types` property should tell you what conversions are possible and you should be able to use it in the same way as any other Data object. Right now it only works for fieldlists and there are also other issues with it.  I am working on an improvement for:

- NetCDF data: to_xarray() and to_fieldlist() should work
- CSV data: to_pandas() should work
- `available_types`: should correctly report the available conversion for any data

Please note that `.sources` was not meant be a public attribute, it is a mistake that is not a private member.

HCookie:
@sandorkertesz 
I was using the `ecfs` source to open a tar archive, and was more wanting to use it to handle the retrieval from that and the local caching. So I'm not sure it would have any `available_types`.

sandorkertesz:
@HCookie, this has been improved. Can I ask you to try the latest `develop` to see if it works for you? What to expect:

- the data object `.path` attribute should work
- the data object `.available_types` attribute should work
- if the data contains csv files `.to_pandas()` should work

If there is still an issue please share the exact retrieval with me.

HCookie:
@sandorkertesz 
The `.path` attribute works nicely now, however I'm not sure the reader is being identified correctly as `tar`

```
[...]

PR #1141, opened 2026-09-10
fix: detect tarfile based on contents

### LINKED ISSUE #676: tar detection is based on mimetypes.guess_type
### What happened?

Fail to detect tar files correctly unless they have correct extension. Should change to base on file content

### What are the steps to reproduce the bug?

-

### Version

-

### Platform (OS and architecture)

-

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

Oisin-M:
### Description
Close #676 

There is a comment
> We don't use tarfile.is_tarfile() because is returns true given a file of zeros

so I have added a check that the file is not all zeroes.

### Contributor Declaration

### DIFF
--- a/src/earthkit/data/readers/tar.py
+++ b/src/earthkit/data/readers/tar.py
@@ -28,7 +28,8 @@ class TarReader(ArchiveReader):
 
 
 def reader(source, path, *, magic=None, deeper_check=False, content_type=None, **kwargs):
-    if tarfile.is_tarfile(path):
+    # we explicitly do not consider all zeroes files as tar
+    if tarfile.is_tarfile(path) and open(path, "rb").read(512).strip(b"\0"):
         return TarReader(source, path, **kwargs)
 
 
--- a/src/earthkit/data/readers/tar.py
+++ b/src/earthkit/data/readers/tar.py
@@ -8,7 +8,6 @@
 #
 
 import logging
-import mimetypes
 import tarfile
 
 from .archive import ArchiveReader
@@ -29,11 +28,7 @@ class TarReader(ArchiveReader):
 
 
 def reader(source, path, *, magic=None, deeper_check=False, content_type=None, **kwargs):
-    # We don't use tarfile.is_tarfile() because is
-    # returns true given a file of zeros
-
-    kind, _ = mimetypes.guess_type(path)
-    if kind == "application/x-tar":
+    if tarfile.is_tarfile(path):
         return TarReader(source, path, **kwargs)
 
 


Oisin-M:
We have a dummy-source zeroes option which can be read as an empty tar instead of presumably a textfile as it was before, and this is used as example files in the cache tests (not sure why this particular choice). I can easily change to use a different dummy source and the tests will pass. However, @sandorkertesz can you clarify what this dummy source creating a `.zeros` file is for? It looks like duplication of the sample source and I question if it is still needed?

sandorkertesz:
> We have a dummy-source zeroes option which can be read as an empty tar instead of presumably a textfile as it was before, and this is used as example files in the cache tests (not sure why this particular choice). I can easily change to use a different dummy source and the tests will pass. However, @sandorkertesz can you clarify what this dummy source creating a `.zeros` file is for? It looks like duplication of the sample source and I question if it is still needed?

A "dummy" source is used for testing only to create inputs from "nothing". The "sample" source is to access hosted example files,  always requires remote access. The name of the "dummy" source is not so great and should be changed. Was planned but never happened.

But the test should pass as it is without any changes in the test itself. "zeros" creates a binary file that was read as "unknown" before, so it should not be read as an empty tar now.

Oisin-M:
> > We have a dummy-source zeroes option which can be read as an empty tar instead of presumably a textfile as it was before, and this is used as example files in the cache tests (not sure why this particular choice). I can easily change to use a different dummy source and the tests will pass. However, @sandorkertesz can you clarify what this dummy source creating a `.zeros` file is for? It looks like duplication of the sample source and I question if it is still needed?
> 
> A "dummy" source is used for testing only to create inputs from "nothing". The "sample" source is to access hosted example files, always requires remote access. The name of the "dummy" source is not so great and should be changed. Was planned but never happened.
> 
> But the test should pass as it is without any changes in the test itself. "zeros" creates a binary file that was read as "unknown" before, so it should not be read as an empty tar now.

Okay, so am I right in understanding that the dummy source should not be part of the public API? I will update the magic checking to handle specifically the zeroes case

Yes, the dummy source is  an undocumented source on  RTD. In that sense it is not part of the public API. But the main point is that  a non-tar file should not be identified as tar.

src/earthkit/data/readers/tar.py
```diff
@@ -29,11 +28,8 @@ def __init__(self, source, path, **kwargs):
 
 
 def reader(source, path, *, magic=None, deeper_check=False, content_type=None, **kwargs):
-    # We don't use tarfile.is_tarfile() because is
-    # returns true given a file of zeros
-
-    kind, _ = mimetypes.guess_type(path)
-    if kind == "application/x-tar":
+    # we explicitly do not consider all zeroes files as tar
+    if tarfile.is_tarfile(path) and open(path, "rb").read(512).strip(b"\0"):
```
sandorkertesz: The value 512 seems to be ad hoc. What if there are more zeros in the file?

sandorkertesz:
@Oisin-M, thanks for this fix. As a follow-up, it would probably be a good idea to add a test that reads a tar file whose name doesn’t have the .tar suffix.

Oisin-M:
> @Oisin-M, thanks for this fix. As a follow-up, it would probably be a good idea to add a test that reads a tar file whose name doesn’t have the .tar suffix.

Good point - have done a small PR here: #1145

sandorkertesz:
@Oisin-M, it was a very useful fix and I just realised it was essential for reading the input data correctly in #1070! Thanks again.

PR #1143, opened 2026-09-15
feat: add pcraster reader

### LINKED ISSUE #1142: Add reader for PCRaster .map format

_No response_

_No response_

_No response_

### Additional context

_No response_

### Organisation

_No response_

Oisin-M:
### Description

Close #1142

Detection unfortunately based on file extension, because there isn't another way to reliably distinguish this format. Added `to_numpy`, no `to_array` added yet

### Contributor Declaration


sandorkertesz:
@Oisin-M, thank you for adding support for this new format. It can be merged once the necessary tests in `tests/readers` and `tests/high_level_object` are added.

Oisin-M:
> @Oisin-M, thank you for adding support for this new format. It can be merged once the necessary tests in `tests/readers` and `tests/high_level_object` are added.

Tests now added for the reader

test.ipynb
```diff
@@ -0,0 +1,113 @@
+{
```
sandorkertesz: This file seems to be a leftover

Oisin-M: Oop yes

src/earthkit/data/data/pcraster.py
```diff
@@ -0,0 +1,78 @@
...
+        except Exception:
+            return None
+
+    def __repr__(self) -> str:
+        return f"PCRasterData(path={self.path})"
+
+    def _repr_html_(self) -> str:
+        return self.describe()._repr_html_()
+
+    def to_numpy(self, mask=True):
```
sandorkertesz, line 65: I wonder if mask is the right name for this kwarg. For me implies 0/1s.

Oisin-M: Happy to change but the meaning here is consistent with the `mask_and_scale` kwarg in `xarray.open_dataset`, albeit without the scale. Could change to `fillna` or similar if preferred?

sandorkertesz: I see. The problem is that `mask` in numpy means something very different (boolean array). I would reject `fillna` since it means turning nans to a specific value.  This kwarg is really a `missing_to_nan`, but it is probably not pythonic enough. I will ask another reviewer to help make the final decision.

iainrussell: I'd go for `missing_to_nan`, as it is very clear. I also wonder if we even need it at all - who would not want to do this conversion?

Oisin-M: Agree, `missing_to_nan` is clear. I have updated. I think it is still be useful to avoid masking in some cases, but I can remove if you think that is better. For example, in earthkit-hydro I currently read without converting to nans some drainage direction codes so I can read into a numpy array with dtype int and not float

sandorkertesz: Thanks for the changes. I think the to_numpy() in Data should also be updated.

Oisin-M: Should be all correct now

